PR #484, opened 2020-06-08
Fix worldborder add showing an error for negative values

### LINKED ISSUE #483: Worldborder add shows an error with negative values
The command `worldborder add n` with n < 0 is correct command syntax, however, the plugin shows an error, saying only values between positive 1 and 60,000,000 are allowed.
This may be true for the set command, the add command supports the range `-60000000..60000000` though (-60000000 will never succeed, since during runtime the size cannot be below 1, it is still correct syntax though and the function file will load)

NeunEinser:
Fixes #483

### DIFF
--- a/contributors/.all-contributorsrc
+++ b/contributors/.all-contributorsrc
@@ -55,6 +55,16 @@
       ]
     },
     {
+      "name": "NeunEinser",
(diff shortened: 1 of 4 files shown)


NeunEinser:
It is actually also correct syntax to have a negative value for set, but I assume this has been changed purposefully, since it doesn't make sense for the set command.

PR #705, opened 2020-09-29
:bug: Fix align code action when not allowed float

### LINKED ISSUE #704: Vector-align-0.0/0.5 is also provided for Vectors where only Integers are allowed

ChenCMD:
Fix #704

### DIFF
--- a/src/test/nodes/VectorNode.spec.ts
+++ b/src/test/nodes/VectorNode.spec.ts
@@ -238,7 +238,7 @@ describe('VectorNode Tests', () => {
             const actual = node[GetCodeActions](uri, info, range, {})
             assert.deepStrictEqual(actual, [])
         })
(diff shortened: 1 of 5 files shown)


src/test/nodes/VectorNode.spec.ts
```diff
@@ -239,5 +238,16 @@ describe('VectorNode Tests', () => {
             const actual = node[GetCodeActions](uri, info, range, {})
             assert.deepStrictEqual(actual, [])
         })
+        it('Should not return align actions when the vector is not allowed float', () => {
```
SPGoding, line 241:
```suggestion
        it('Should not return align actions when the vector does not allow float', () => {
```

SPGoding (approved):

SPGoding (approved):

PR #731, opened 2020-11-05
Update json localization

misode:
Required changes for to be able to upgrade `@mcschema/core` to `0.11.0`.

The language server now needs to handle localization itself since it has been removed from @mcschema/core. But since it already had localization I decided to simply merge the objects. *I assume there are no conflicting translation keys?*

This also adds a `segmentedLocale` exported function in `locales/index.ts`. This is the equivalent of `Path.strictLocale`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,8 +31,8 @@
         "watch": "tsc -w"
     },
     "dependencies": {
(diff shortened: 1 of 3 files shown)


SPGoding (approved):
I LOVE YOU!

PR #741, opened 2020-11-12
:recycle: Suggestions to replace the extname function

ChenCMD:
path.extname() should not be used for files with only an extension, as it will not be recognized as an extension.
There is currently no implementation of referencing ext in IdentityNode.fromRel(), but this is to prevent future bugs.
refer: https://nodejs.org/api/path.html#path_path_extname_path


SPGoding:
Removed `ext` instead

PR #755, opened 2020-11-13
:bug: Fix doesn't work nameOf*

### LINKED ISSUE #750: nameOf* other than nameOfNbtCompoundTagKeys does not work
image

ChenCMD:
Fix #750
image


src/parsers/TeamArgumentParser.ts
```diff
@@ -50,6 +52,18 @@ export class TeamArgumentParser extends ArgumentParser<string> {
             ))
         } else {
             if (this.isDefinition) {
```
SPGoding, line 54: Same here.

src/parsers/ObjectiveArgumentParser.ts
```diff
@@ -50,6 +52,18 @@ export class ObjectiveArgumentParser extends ArgumentParser<string> {
             ))
         } else {
             if (this.isDefinition) {
```
SPGoding, line 54: Any reason why the check is limited to definitions?

ChenCMD:
I thought everyone would define it and then use the scoreboard and teams, which in hindsight is a silly idea.

SPGoding (approved):

PR #799, opened 2020-12-09
Fix validation of value, score, nbt providers

misode:


SPGoding (approved):
wtf happened to the line changes lol

PR #877, opened 2021-04-02
Add Completer and JSON completion

misode:


packages/core/src/processor/util.ts
```diff
@@ -58,3 +58,22 @@ export function selectedLeaf(node: AstNode, offset: number, existingParents: Ast
 	}
 	return null
 }
+
+export function selectedNode(node: AstNode, offset: number): { node: AstNode, parents: AstNode[] } | null
+export function selectedNode(node: AstNode, offset: number, existingParents: AstNode[] = []): { node: AstNode, parents: AstNode[] } | null {
```
misode, line 63: This is mostly a copy of `selectedLeaf`, but instead of only returning leaves it always returns the deepest node. It's possible these two could be merged.

packages/core/src/processor/Completer.ts
```diff
@@ -0,0 +1,57 @@
...
+export interface CompletionToken {
+	label: string,
+	kind?: CompletionKind,
+	detail?: string,
+	documentation?: string,
+	deprecated?: boolean,
+	insertText?: string,
+	sortText?: string,
+	filterText?: string,
+}
```
misode, line 48: This is very similar to the LSP `CompletionItem` type. I'm not sure how much should be copied vs abstracted.

packages/core/src/processor/util.ts
SPGoding: What's the difference between leaf and the deepest node :eyes:

SPGoding: Ah I just saw the difference. My old `selectedLeaf` made an assumption that the `children` of a node covers every character of this node, which is fixed by your `selectedNode` implementation. Yeah I think you could just replace the body of `selectedLeaf` with the one of `selectedNode`; that's how it is supposed to work.

SPGoding (approved):

PR #908, opened 2021-10-29
WIP :sparkles: Allow Chikidar options to be passed

strootje:
updated project to accept (optionally) chokidar options


SPGoding:
Sorry for getting back so late -- I would like to keep chokidar as implementation details and not expose its options in the API. Let me know if you have any use cases for custom watcher options.

PR #1047, opened 2023-05-07
Small fixes in mcdoc examples

vberlier:
I was implementing a parser and noticed that some of the examples were invalid

### DIFF
--- a/docs/user/mcdoc/index.adoc
+++ b/docs/user/mcdoc/index.adoc
@@ -1001,12 +1001,12 @@ which will get replaced by actual types when the type alias is instantiated else
 type NumericRange<T> = ( <1>
 	T | <2>
 	[T, T] | <2>
-	struct { min: T, max: T }, <2>
+	struct { min: T, max: T } <2>
 )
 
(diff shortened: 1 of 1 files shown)


docs/user/mcdoc/index.adoc
```diff
@@ -1001,12 +1001,12 @@ which will get replaced by actual types when the type alias is instantiated else
 type NumericRange<T> = ( <1>
 	T | <2>
 	[T, T] | <2>
-	struct { min: T, max: T }, <2>
+	struct { min: T, max: T } <2>
```
misode, line 1004: Trailing commas are allowed and even recommend in vanilla-mcdoc, so I'd leave this

vberlier: Isn't this is a union type though? A trailing pipe would be valid but I'm not sure about a trailing comma.

misode: Oh good point yes

misode (approved):

PR #1077, opened 2023-10-01
Update DHP 3 command format to 1.20+

Trivaxy:
The latest versions of Minecraft brought a lot of new features. This WIP PR targets one big part - the command syntax changes - by:
- Adding function macro support by making macro lines valid syntax components. DHP will:
    - Check that the macro line syntax is correct
    - Check that at least one macro variable is used since vanilla doesn't allow macros with no variables
    - Provide autocompletion for macro variables by keeping track of what variables were written previously in the file
    - Maybe an option to allow the user to expand a macro and preview what it would look like?

- Adding the `random` command
    - self-explanatory
    
- Updating `function` call syntax to account for function macros. DHP will:
    - Allow passing in NBT compounds directly, or via `with` syntax
    - Raise errors when a function macro is not passed any arguments
    - Raise errors when a function macro is passed a literal without the proper arguments
    - Raise a warning when a function macro is passed a literal with unnecessary arguments
    
- Adding support for `\` line continuations
    - Hopefully it's not too bad to implement...

This is only one part of the update since 1.20 adds a decent amount of data-related changes and additions. Not sure if I want to tackle them


src/plugins/builtin/McfunctionPlugin.ts
```diff
@@ -12,40 +12,152 @@ export class McfunctionPlugin implements plugins.Plugin {
...
+            ans.range.end = macroReader.cursor
+        }
+        catch (p) {
+            ans.errors.push(p)
+        }
+
+        if (!hasVariables) {
+            ans.errors.push(new ParsingError(
+                { start: ans.range.start, end: ans.range.end },
+                'Must have at least one macro variable',
```
misode, line 82: Use a localized string here

Trivaxy: Forgot to do that. Will do shortly

misode (approved):

Issue #1078, opened 2023-10-11
below_name returns an error


InQuognito:
![image](https://github.com/SpyglassMC/Spyglass/assets/44120299/fea2faf0-c418-4243-b5c4-37215413577d)

Pretty simple.

PR #1079, opened 2023-10-12
Allow below_name as a display slot for 1.20.2

Trivaxy:
Fixes https://github.com/SpyglassMC/Spyglass/issues/1078.

DHP will now recognize `below_name` for 1.20.2 and above, and use `belowName` for older versions.


ChenCMD (approved):
LGTM
NITS: If less than 1.20.2, it would be even better to have a test to check that the belowName is interpreted correctly.

misode:
Works, thank you!

PR #1080, opened 2023-10-14
:bug: Fix macro indentation not being taken into account

### LINKED ISSUE #1081: Indexed macro lines give error warning
When a macro line gets indented (by any amount) it gives this error "Failed to match any children in the command syntax tree".
If the indentation is deleted the error goes away

ChenCMD:
Fix #1081

### DIFF
--- a/src/plugins/builtin/McfunctionPlugin.ts
+++ b/src/plugins/builtin/McfunctionPlugin.ts
@@ -29,7 +29,7 @@ export class CommandSyntaxComponentParser implements plugins.SyntaxComponentPars
     identity = 'spgoding:mcfunction/command'
 
     test(reader: StringReader, ctx: ParsingContext): [boolean, number] {
-        return [reader.peek() !== '$', 0]
+        return [reader.skipSpace().peek() !== '$', 0]
     }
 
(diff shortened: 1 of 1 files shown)


misode:
Can you give an example command where this issue applies?

ChenCMD:
```mcfunction
# macro command with indent
    $say $(message)
```

SPGoding (commented):
Haven't touched the code for a while but I think the `test` methods are not supposed to mutate the state of the `reader`. And iirc the `skipSpace()` method will change the cursor of it. Not sure how to fix that though.

ChenCMD:
> Haven't touched the code for a while but I think the `test` methods are not supposed to mutate the state of the `reader`. And iirc the `skipSpace()` method will change the cursor of it. Not sure how to fix that though.
> @SPGoding 


Since the StringReader is cloned and passed to SyntaxComponentParser#test, changing the cursor doesn't seem to be a problem, does it?
https://github.com/SpyglassMC/Spyglass/blob/1a18e7eeb3367d3c44157d979194658d5243fff8/src/services/common.ts#L108

SPGoding (approved):
Yeah you're right, sorry!

PR #1091, opened 2023-12-15
:bug: Support 1.20.3

ChenCMD:


EstEarth202:
@ChenCMD Eh? I would like to ask first to be sure, does this update have a command `return fail` as well?

ChenCMD:
@EstEarth202 
> @ChenCMD Eh? I would like to ask first to be sure, does this update have a command `return fail` as well?

Oh, I didn't know there was such a thing. I'll add it.

EstEarth202:
Glad to help. Because I remember from slicedlime video that it was added.

> A new subcommand `return fail` has been added to make the whole function fail (i.e. return success=0 and result=0)

But I saw it says subcommand so I'm not sure.

misode:
Change look good in general, will make adjustments after merging

PR #1119, opened 2024-05-05
:no_entry_sign: Specify Node 16 in tsconfig.json

Calverin:
In `Spyglass\packages\core\src\tsconfig.json` the module should be set to Node 16 to avoid issues when trying to build the extension.


MulverineX:
Nothing should specify Node16, this should be solved in another manner

Issue #1069, opened 2023-08-05
Support backslash continuation in functions


SPGoding:
> * A single backslash `\` as the last non-whitespace character of a line now allows a command to be continued on the next line
>     * Leading and trailing whitespace of the following line are stripped before appending
>
> -- https://www.minecraft.net/en-us/article/minecraft-snapshot-23w31a

Potential solutions:

1. Change every command parser to support backslash continuation
2. Write a wrapper parser to concatenate backslash-continued commands, and parse the concatenated command with the current command parser. An `IndexMap` should be used to keep track of the index mappings.

PR #1124, opened 2024-05-08
🐛 Fix `IndexMap`'s `toInnerRange` and `toOuterRange` returning extended ranges

TheAfroOfDoom:
ran into this while working on #1069 

we were lacking in unit test coverage when it came to how `IndexMap`s interact with `Source`s, so the first commit adds some to make this error stand out more obviously: [`24c5607` (#1124)](https://github.com/SpyglassMC/Spyglass/pull/1124/commits/24c5607ee3cc6c9117ac80e7edc79230512be0b4)

see the following example (same as the new unit tests):

```ts
/*
 * Index Tens - 000000000011111111112
 * Index Ones - 012345678901234567890
 * Outer      - foo\"bar\u00a7qux
 * Inner      - foo"bar§qux
 */

const indexMap = [
	// `"` to `\"`
	{ inner: Range.create(3, 4), outer: Range.create(3, 5) },
	// `§` to `\u00a7`
	{ inner: Range.create(7, 8), outer: Range.create(8, 14) },
]
```

trying to fetch the **outer** range of `r` in the above case _should_ output `[7, 8)` (`r` is shifted 1 index to the right)

instead `r`'s outer range was being calculated as `[7, 14)` since `8)` overlaps with `[8, 14)` **when `IndexMap.convertOffset` is given the flag `isEndOffset: true`

---

all tests pass after this change, but there _might_ be unintended side effects? i couldn't figure out the original intention behind the addition of the `isEndOffset` flag

would appreciate @misode 's input on that ^

---

this PR also has some miscellaneous semi-related cleanup

### DIFF
--- a/packages/core/src/source/Source.ts
+++ b/packages/core/src/source/Source.ts
@@ -313,24 +313,26 @@ export class Source extends ReadonlySource {
 		this.readRemaining()
 		return this
 	}
+}
 
-	static isDigit(c: string): c is Digit {
+export namespace Source {
+	export function isDigit(c: string): c is Digit {
 		return c >= '0' && c <= '9'
 	}
 
-	static isBrigadierQuote(c: string): c is '"' | "'" {
+	export function isBrigadierQuote(c: string): c is '"' | "'" {
 		return c === '"' || c === "'"
 	}
 
-	static isNewline(c: string): c is Newline {
+	export function isNewline(c: string): c is Newline {
 		return c === '\r\n' || c === '\r' || c === '\n'
 	}
 
-	static isSpace(c: string): c is Space {
+	export function isSpace(c: string): c is Space {
 		return c === ' ' || c === '\t'
 	}
 
-	static isWhitespace(c: string): c is Whitespace {
+	export function isWhitespace(c: string): c is Whitespace {
 		return Source.isSpace(c) || Source.isNewline(c)
 	}
 }
--- a/packages/core/src/source/Source.ts
+++ b/packages/core/src/source/Source.ts
@@ -110,10 +110,10 @@ export class ReadonlySource {
 			cursor++
 		) {
 			const c = this.string.charAt(cursor)
-			if (c === CR || c === LF) {
+			if (Source.isNewline(c)) {
 				break
 			}
-			if (!(c === ' ' || c === '\t')) {
+			if (!Source.isSpace(c)) {
 				return true
 			}
 		}
--- a/packages/core/src/source/Source.ts
+++ b/packages/core/src/source/Source.ts
@@ -176,7 +176,7 @@ export class Source extends ReadonlySource {
 
 	/**
 	 * Skips the current character.
-	 * @param step The step to skip. @default 1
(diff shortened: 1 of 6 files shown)


SPGoding:
`isEndOffset` parameter is added in c84d6ae5d1fe94a01f6c72d127ba4563b4336af4 along with test cases for `IndexMap.merge()`. Those tests are still passing so i guess it's not impacted? Not sure why i added that parameter in that case.

TheAfroOfDoom:
I think I have an idea of cases that may break under the current implementation, will look into that today

TheAfroOfDoom:
> I think I have an idea of cases that may break under the current implementation

the only semi-strange case was an outer range of `r\u00` being converted to an inner range containing `r`. not sure if we would prefer the other behavior (of inner being `r§`)

packages/core/src/source/Source.ts
```diff
@@ -313,24 +313,26 @@ export class Source extends ReadonlySource {
 		this.readRemaining()
 		return this
 	}
+}
 
-	static isDigit(c: string): c is Digit {
+export namespace Source {
+	export function isDigit(c: string): c is Digit {
```
misode, line 319: What's the reason for switching to a namespace instead of static methods?

misode (approved):
Very minimal changes and seems to clean up the logic. Since the tests are still all passing this seems safe to merge.

TheAfroOfDoom:
@misode from [discord](https://discord.com/channels/666020457568403505/666037123450929163/1238239921651712030):

> Do you have a concrete example with line continuations were it screwed up before?

i've been working at finding a concrete example of incorrect range highlighting due to this bug but ran into various other bugs blocking me from doing so this week (including the "expected a string or a string")

the only place this would surface is where we use `Source.getCharRange()`

`Source.getCharRange()` is only called in the string parser:

https://github.com/SpyglassMC/Spyglass/blob/b6ecdcad78be7a44ef564afdfc9c51daedf38de8/packages/core/src/parser/string.ts#L59-L63

https://github.com/SpyglassMC/Spyglass/blob/b6ecdcad78be7a44ef564afdfc9c51daedf38de8/packages/core/src/parser/string.ts#L75-L81

from my message [here](https://discord.com/channels/666020457568403505/666037123450929163/1239074032255762512) though, i dont think these unicode/escape errors are actually surfacing in vscode even with the right contexts, so its hard to actually show #1124's bug

---

imo its fine to merge this PR before i'm able to resolve the above issue since this behavior is well documented with unit tests

packages/core/src/source/Source.ts
TheAfroOfDoom: i thought namespaces more "cleanly" specified the intention behind these static functions. i share your question you left here: https://github.com/SpyglassMC/Spyglass/pull/1129#discussion_r1597703484

i think namespaces look cleaner than static functions but am not super committed either way. it'd be nice if we chose one and stuck with it though

i don't think theres a functional difference?

SPGoding (approved):

PR #1127, opened 2024-05-11
🔥 Remove old json checkers

misode:
No need to have this old system confusing us when it's going to be replaced by mcdoc.


TheAfroOfDoom (changes requested):
can we also remove the snapshot code that is generated by tests that used `testGrid` / `testChecker`?

i don't know if theres a more official way of updating snapshots but ive just been deleting the relevant file(s) and re-running `npm run test` (this is more of a "how do you do it?" question than a "this is how you should do it")

SPGoding:
Setting the environment variable `SNAPSHOT_UPDATE=1` when running `npm test` should update the snapshot but i just delete the relevant snapshot files too lol

SPGoding (approved):

PR #1129, opened 2024-05-11
Improve AST node type guards

SPGoding:
Renamed the utility type `NodeIsHelper` to ~~`PotentiallyReadonly`~~ `InheritReadonly` and improved documentation and tests.

### DIFF
--- a/packages/core/src/common/ReadonlyProxy.ts
+++ b/packages/core/src/common/ReadonlyProxy.ts
@@ -1,8 +1,12 @@
 import { emplaceMap, isObject } from './util.js'
 
-type Wrap<T> = T extends object ? DeepReadonly<T> : T
+type DeepReadonlyValue<T> = T extends object ? DeepReadonly<T> : T
 export type DeepReadonly<T extends object> = {
-	readonly [K in keyof T]: Wrap<T[K]>
+	readonly [K in keyof T]: DeepReadonlyValue<T[K]>
+}
+
+export type ReadWrite<T extends object> = {
+	-readonly [K in keyof T]: T[K]
 }
 
 export const ReadonlyProxy = Object.freeze({
--- a/packages/core/src/common/ReadonlyProxy.ts
+++ b/packages/core/src/common/ReadonlyProxy.ts
@@ -9,11 +9,11 @@ export type ReadWrite<T extends object> = {
 	-readonly [K in keyof T]: T[K]
 }
 
-export const ReadonlyProxy = Object.freeze({
-	create<T extends object>(obj: T): DeepReadonly<T> {
+export namespace ReadonlyProxy {
+	export function create<T extends object>(obj: T): DeepReadonly<T> {
 		return new Proxy(obj, new ReadonlyProxyHandler()) as any
-	},
-})
+	}
+}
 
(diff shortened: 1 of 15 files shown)


TheAfroOfDoom (approved):
what `PotentiallyReadonly` does makes enough sense to me, do you have an example of when/where we'd want to use it practically? (/ already do)

SPGoding:
> Do you have an example of when/where we'd want to use it practically? (/ already do)

Ideally it should be used in every type guard for an AST node. Some processors (like completers) take a `DeepReadonly<>` form of the node while other things (like checkers) take the mutable form, so using `PotentiallyReadonly<>` in the node's type guards would make sure we don't lose the readonly information inside the body of completers etc.

packages/core/src/common/ReadonlyProxy.ts
```diff
@@ -1,8 +1,12 @@
...
 
-type Wrap<T> = T extends object ? DeepReadonly<T> : T
+type DeepReadonlyValue<T> = T extends object ? DeepReadonly<T> : T
 export type DeepReadonly<T extends object> = {
-	readonly [K in keyof T]: Wrap<T[K]>
+	readonly [K in keyof T]: DeepReadonlyValue<T[K]>
+}
+
+export type ReadWrite<T extends object> = {
+	-readonly [K in keyof T]: T[K]
```
misode, line 9: Is it intended that `ReadWrite` is not deep, in contrast to `DeepReadonly`?

packages/mcfunction/src/node/command.ts
```diff
@@ -26,7 +26,7 @@ export interface CommandMacroNode extends core.AstNode {
 export const CommandMacroNode = Object.freeze({
 	is<T extends core.DeepReadonly<core.AstNode> | undefined>(
 		obj: T,
-	): obj is core.NodeIsHelper<CommandMacroNode, T> {
+	): obj is core.PotentiallyReadonly<CommandMacroNode, T> {
```
misode, line 29: When should we use namespaces and when `Object.freeze`? What is the difference?

SPGoding: I only stopped using `namespace` since it's a TS-specific thing instead of a JS thing. We should decide on a style and keep using that.

packages/core/src/common/ReadonlyProxy.ts
SPGoding: yes.

SPGoding:
Also potential bike-shedding on the type name `PotentiallyReadonly`, would `InheritReadonly` be a more fitting name, since that the type guard does is basically inheriting the readonly-ness of `INPUT` on `TARGET`?

misode:
> Also potential bike-shedding on the type name `PotentiallyReadonly`: would `InheritReadonly` be a more fitting name, since what the type does is basically inheriting the readonly-ness of `INPUT` on `TARGET`?

That sounds like a good name yes!

SPGoding:
I changed the name to `InheritReadonly` and changed all `Object.freeze()`s to `namespace`s.

MulverineX (approved):

Issue #1118, opened 2024-05-05
Item Component Support


Calverin:
Support for error checking and autocomplete for item components

misode:
I removed the mcdoc parts of the description, that can be a separate issue.

PR #1131, opened 2024-05-13
✨ Add support for item components

Trivaxy:
This PR lays the foundation for supporting the new Item Component format Mojang introduced in `1.20.5`.

There is still work to be done - for now, this PR is a draft but almost ready for merging.

# Done
- [x] Implement parsing for item components, e.g. `give @s cooked_beef[food={nutrition:5,saturation:2}]`
- [x] Validate and throw errors when duplicate components are defined, e.g. `give @s stick[food=...,food=...]`
- [x] Implement support for the new item predicate format, used in `execute if|unless` and `clear`

# What's left (outside this PR's scope)
- [ ] Validate the values given to components. This is waiting on mcdoc (will be done in a separate PR)
- [ ] Validate JSON files which use item components. This is waiting on mcdoc (will be done in a separate PR)


TheAfroOfDoom:
> [ ] Validate the values given to components. This is waiting on mcdoc (will be done in a separate PR)
> [ ] Validate JSON files which use item components. This is waiting on mcdoc (will be done in a separate PR)

these could be spun into separate tickets maybe? or at least once this PR is merged we should update the original ticket https://github.com/SpyglassMC/Spyglass/issues/1118 with the progress of item components and what needs to still be done, etc.

MulverineX (approved):

packages/java-edition/src/mcfunction/completer/argument.ts
```diff
@@ -56,9 +59,17 @@ import {
 } from '../node/index.js'
 import type { ArgumentTreeNode } from '../tree/index.js'
 
+function getItemStackFormat(ctx: CompleterContext): 'old' | 'new' {
+	const release = ctx.config.env.gameVersion as ReleaseVersion | undefined
```
NeunEinser, line 63: What's the difference between this and `ctx.project['loadedVersion]`?

SPGoding: I think `ctx.project['loadedVersion]` should be used here? `ctx.config.env.gameVersion` is the raw config value the user provides (it could be something weird like `'auto'` and `'latest snapshot'`) while `ctx.project['loadedVersion]` is an actual game version we parsed

packages/locales/src/locales/en.json
```diff
@@ -33,6 +33,7 @@
   "conjunction.or_3+_1": ", ",
   "conjunction.or_3+_2": ", or ",
   "datafix.error.command-replaceitem": "/replaceitem was removed in 20w46a (the second snapshot of 1.17) in favour of /item",
+  "duplicate-components": "Duplicate component definitions for %0%",
```
SPGoding, line 36: Rename the key to `mcfunction.parser.item.duplicate-components` and put it with the other `mcfunction` error messages in the file.

packages/java-edition/src/mcfunction/checker/index.ts
```diff
@@ -162,14 +166,46 @@ const entity: core.SyncChecker<EntityNode> = (node, ctx) => {
...
+		groupedComponents.forEach((components, componentName) => {
+			if (components.length > 1) {
+				components.forEach(component => {
+					ctx.err.report(
+						localize('duplicate-components', componentName),
+						component.key!.range,
+						core.ErrorSeverity.Error,
+					)
+				})
+			}
```
SPGoding, line 206: You can use `core.LanguageErrorInfo.related` to report only one error for this with the duplicated keys' locations added in the `info.related` field (`core.Location.create()` would be helpful to create the entries for the field)

packages/java-edition/src/mcfunction/node/argument.ts
```diff
@@ -304,6 +349,61 @@ export namespace ItemNode {
...
+
+export namespace ComponentListNode {
+	export function is(node: core.AstNode): node is ComponentListNode {
+		return (node as ComponentListNode).type === 'mcfunction:component_list'
+	}
+}
+
+export interface ComponentPredicatesNode extends core.AstNode {
+	type: 'mcfunction:component_predicates'
+	children: ComponentTestBaseNode[]
```
SPGoding, line 365: Instead of `ComponentTestBaseNode` I think you'd want to create a new union type `type ComponentTestNode = ComponentTestExactNode | ComponentTestExistsNode | ComponentTestSubpredicateNode` and use that here as the children.

packages/java-edition/src/mcfunction/node/argument.ts
```diff
@@ -304,6 +349,61 @@ export namespace ItemNode {
...
+}
+
+export interface ComponentTestExistsNode extends ComponentTestBaseNode {
+	type: 'mcfunction:component_test_exists'
+	children: [core.ResourceLocationNode]
+	component: core.ResourceLocationNode
+}
+
+export interface ComponentTestSubpredicateNode extends ComponentTestBaseNode {
+	type: 'mcfunction:component_test_subpredicate'
```
SPGoding, line 401: (minor) "sub predicate" are two words

```suggestion
export interface ComponentTestSubPredicateNode extends ComponentTestBaseNode {
	type: 'mcfunction:component_test_sub_predicate'
```

packages/java-edition/src/mcfunction/node/argument.ts
```diff
@@ -304,6 +349,61 @@ export namespace ItemNode {
...
+	children: [core.ResourceLocationNode, nbt.NbtNode]
+	component: core.ResourceLocationNode
```
SPGoding, line 403: The resource location does not identify a component here, it identifies a sub predicate instead.
```suggestion
	subPredicateType: core.ResourceLocationNode
```

packages/java-edition/src/mcfunction/node/argument.ts
```diff
@@ -304,6 +349,61 @@ export namespace ItemNode {
...
+	subpredicate: nbt.NbtNode
```
SPGoding, line 404: (minor)

```suggestion
	subPredicate: nbt.NbtNode
```

packages/java-edition/src/mcfunction/node/argument.ts
```diff
@@ -279,22 +279,67 @@ export interface FloatRangeNode extends core.AstNode {
...
+
+export interface ItemNewNode extends core.AstNode {
+	type: 'mcfunction:item'
+	children: (
+		| core.LiteralNode
+		| core.ResourceLocationNode
+		| ComponentListNode
+		| ComponentPredicatesNode
+	)[]
+	id: core.ResourceLocationNode
```
SPGoding, line 303: If this is a wildcard item predicate node, `id` would be `undefined`, so it has to be optional. I wonder if it makes sense to change this to `id: core.ResourceLocationNode | core.LiteralNode` and remove `wildcard?: boolean`: by writing the type that way we are guaranteed one and only one of them will exist and we can check its type to know if it's wildcard.

packages/java-edition/src/mcfunction/node/argument.ts
```diff
@@ -279,22 +279,67 @@ export interface FloatRangeNode extends core.AstNode {
...
+	components?: ComponentListNode
+	componentPredicates?: ComponentPredicatesNode
+	wildcard?: boolean
+}
+
+export namespace ItemNewNode {
+	export function is(node: core.AstNode | undefined): node is ItemNewNode {
+		return (node as ItemNewNode | undefined)?.type === 'mcfunction:item' &&
+			'components' in (node as ItemNode)
```
SPGoding, line 312: It this is a item predicate node `components` won't exist, so this check won't work as intended. This shouldn't be needed if we separate the `type` for `ItemOldNode` and `ItemNewNode` though as mentioned in earlier comment.

packages/java-edition/src/mcfunction/parser/argument.ts
```diff
@@ -496,8 +520,75 @@ function item(isPredicate: boolean): core.InfallibleParser<ItemNode> {
...
+					core.resourceLocation({ category: 'item', allowTag: true }),
+				]),
+				core.optional(core.failOnEmpty(componentTests())),
+			]),
+			(res) => {
+				const ans: ItemNewNode = {
+					type: 'mcfunction:item',
+					range: res.range,
+					children: res.children,
+					id: res.children.find(core.ResourceLocationNode.is)!,
```
SPGoding, line 550: This non-null assertion is not safe.

packages/java-edition/src/mcfunction/parser/argument.ts
```diff
@@ -1640,3 +1731,173 @@ function vector(
...
+	const subpredicate = core.map<
+		core.SequenceUtil<
+			core.ResourceLocationNode | core.LiteralNode | nbt.NbtNode
+		>,
+		ComponentTestSubpredicateNode
+	>(
+		core.sequence([
+			core.resourceLocation({
+				pool: [],
+				allowUnknown: true,
```
SPGoding, line 1827: This should use the `item_sub_predicate_type` registry (already exists in mcmeta registries, just needs to add to `RegistryCategories`).

packages/java-edition/src/mcfunction/parser/argument.ts
```diff
@@ -1640,3 +1731,173 @@ function vector(
...
+		},
+	])
+
+	return core.map(
+		core.list({
+			start: '[',
+			value: test,
+			sep: [',', '|'],
+			end: ']',
+			trailingSep: true,
```
SPGoding, line 1891: This would allow trailing `|`'s, although upon my testing trailing commas aren't allowed by the game in item predicates (but they are allowed in non predicate), so just removing `trailingSep: true` probably works (would be nice if a comment could be added there to explain this). The parse tree also won't reflect the real hierarchy of the tests but I guess that doesn't matter much for us.

packages/java-edition/src/mcfunction/parser/argument.ts
```diff
@@ -1640,3 +1731,173 @@ function vector(
...
+function components(): core.InfallibleParser<ComponentListNode> {
+	return core.map(
+		core.record({
+			start: '[',
+			pair: {
+				// TODO: Create an item_component category
+				key: core.resourceLocation({
+					pool: [],
+					allowUnknown: true,
+					namespacePathSep: '.',
```
SPGoding, line 1744: Why is `namespacePathSep` set to `.`? I think the game uses `:` to separate namespace and path here.

SPGoding (commented):
Can you merge the `main` branch in and add some tests in `packages/java-edition/test/mcfunction/parser/argument.spec.ts` to document the parse tree? You might need to add some logic in the test file so that both the old item parser and the new item parser can be tested.

packages/java-edition/src/mcfunction/parser/argument.ts
Trivaxy: The parse tree you see barely functions as is, unfortunately. There's (supposed to) be a push to fix it when I have the time

Trivaxy:
> Can you merge the `main` branch in and add some tests in `packages/java-edition/test/mcfunction/parser/argument.spec.ts` to document the parse tree? You might need to add some logic in the test file so that both the old item parser and the new item parser can be tested.

Sure, once the parser is fixed and the PR is ready to incorporate all the reviews

packages/java-edition/src/mcfunction/checker/index.ts
misode: Vscode also creates an error for each duplicate key in a json file for example, imo this is the most clean

misode:
I did some refactoring. It is mostly done now apart from support for `|` and the completer for item predicate components.

Trivaxy:
The PR should be ready for final reviews now

packages/java-edition/src/mcfunction/node/argument.ts
```diff
@@ -293,31 +293,155 @@ export interface FloatRangeNode extends core.AstNode {
...
 			range: core.Range.get(range),
 			children: [id],
 			id,
 		}
 	}
 }
 
+export interface ComponentTestsNode extends core.AstNode {
+	type: 'mcfunction:component_tests'
+	children: ComponentTestsAnyOfNode[]
```
misode, line 372: I think there is a layer of children too much here. This `children` field will always have exactly zero or one child. Maybe the best way to solve this is by getting rid of this node, and using `ComponentTestsAnyOfNode` in `ItemPredicateNode`

packages/java-edition/src/mcfunction/parser/argument.ts
```diff
@@ -1767,3 +1822,187 @@ function vector(
...
+	)
+	src.skipWhitespace()
+
+	if (src.trySkip('=')) {
+		const value = nbt.parser.entry(src, ctx)
+
+		if (value == core.Failure) {
+			ctx.err.report(localize('expected', localize('nbt.node')), src)
+			src.skipUntilOrEnd(',', '|', ']')
+			return core.Failure
```
misode, line 1866: This should not return a failure. Since you already reported an error, it should return a `ComponenttestExactNode` but with `value` undefined.

packages/java-edition/src/mcfunction/parser/argument.ts
```diff
@@ -1767,3 +1822,187 @@ function vector(
...
+		return ans
+	}
+	if (src.trySkip('~')) {
+		resLoc.options.category = 'item_sub_predicate_type'
+		const predicate = nbt.parser.entry(src, ctx)
+
+		if (predicate == core.Failure) {
+			ctx.err.report(localize('expected', localize('nbt.node')), src)
+			src.skipUntilOrEnd(',', '|', ']')
+			return core.Failure
```
misode, line 1887: Same story here.

misode (changes requested):

packages/java-edition/src/mcfunction/node/argument.ts
Trivaxy: Originally `ComponentTestsAnyOfNode` was the one being returned, but IMO encapsulating it in a `ComponentTestsNode` is more readable and maybe less prone to breakage if the representation for the AST changes

misode: I disagree with that. We don't need unnecessary wrapper nodes. It also makes things like checkers and completers more complex for no reason.

packages/java-edition/src/mcfunction/parser/argument.ts
Trivaxy: See 5cfdfa1fb9b8d9c393409a2c1252e7ad42f7b458

packages/java-edition/src/mcfunction/parser/argument.ts
Trivaxy: See 5cfdfa1fb9b8d9c393409a2c1252e7ad42f7b458

packages/java-edition/src/mcfunction/node/argument.ts
Trivaxy: Hmm I'm not so sure. My main concern is future-proofing, since having things refer to `ComponentTestsNode` seems better than having to change all references from `ComponentTestsAnyOfNode` to `MaybeSomethingMoreComplicatedNode[]` if that ever happens. I doubt it's a big deal in either case so I'm OK with the change.

misode (approved):

PR #1132, opened 2024-05-13
🐛 Fix JSON array/object parsers not absorbing child parser errors

TheAfroOfDoom:
# Summary

these parsers were not absorbing their child parser's errors due to them calling the JSON `entry` parser (`json(dumpErrors = true)`)

with `dumpErrors = true`, the child parsers would dump errors into a temporary context that wasn't actually then sent to the language server

https://github.com/TheAfroOfDoom/Spyglass/blob/b6ecdcad78be7a44ef564afdfc9c51daedf38de8/packages/json/src/parser/entry.ts#L40-L42

fix is to not use the `entry` parser and instead use `json(dumpErrors = false)` so the errors from child parsers are absorbed later by the parent parsers and ultimately outputted to the language server

https://github.com/TheAfroOfDoom/Spyglass/blob/b6ecdcad78be7a44ef564afdfc9c51daedf38de8/packages/core/src/parser/util.ts#L67-L70

the JSON `array` and `object` parsers were the only two places i found that were incorrectly using the `entry` parser

---

### JSON `array` parser

| before | after |
|-|-|
| image | image |

### JSON `object` parser

| before | after |
|-|-|
|image| image|

## Blockers

~~this PR likely should wait until https://github.com/SpyglassMC/Spyglass/pull/1127 is merged since that PR fixes a related bug~~

~~(the branch is currently rebased off of #1127, so those commits will show up here)~~

#1127 has been merged, so no more blockers 😄

### DIFF
--- a/packages/core/src/parser/util.ts
+++ b/packages/core/src/parser/util.ts
@@ -560,3 +560,14 @@ export function acceptIf<P extends Parser<AstNode>>(
 		return ans
 	}) as ReturnType<typeof acceptIf<P>>
 }
+
+/**
+ * @returns A parser that dumps any parser errors after it finishes parsing.
+ */
+export function dumpErrors<P extends Parser<AstNode>>(parser: P): P {
+	return ((src: Source, ctx: ParserContext) => {
+		const ans = parser(src, ctx)
+		ctx.err.dump()
+		return ans
+	}) as ReturnType<typeof dumpErrors<P>>
+}
--- a/packages/core/test/parser/util.spec.ts
+++ b/packages/core/test/parser/util.spec.ts
@@ -1,3 +1,4 @@
+import assert from 'assert'
 import { describe, it } from 'mocha'
 import snapshot from 'snap-shot-it'
 import type {
@@ -7,7 +8,7 @@ import type {
 	Result,
 	Source,
 } from '../../lib/index.js'
-import { any, Failure, Range } from '../../lib/index.js'
+import { any, boolean, dumpErrors, Failure, Range } from '../../lib/index.js'
 import { showWhitespaceGlyph, testParser } from '../utils.js'
 
 interface LiteralNode extends AstNode {
@@ -93,6 +94,42 @@ describe('core util parsers', () => {
 					snapshot(testParser(parser, content))
 				},
 			)
+		}
+	})
+
+	describe('dumpErrors()', () => {
+		const suites: {
+			name: string
+			parser: Parser<AstNode>
+			content: string
+			expectedErrors: any[]
+		}[] = [
+			{
+				name: 'should output errors when not wrapped with `dumpErrors()`',
+				parser: boolean,
+				content: 'bar',
+				expectedErrors: [{
+					range: {
+						start: 0,
+						end: 0,
+					},
+					message: 'Expected “false” or “true”',
+					severity: 3,
+				}],
+			},
+			{
+				name: 'should not output errors when wrapped with `dumpErrors()`',
+				parser: dumpErrors(boolean),
+				content: 'bar',
+				expectedErrors: [],
+			},
+		]
+		for (const { name, content, parser, expectedErrors } of suites) {
+			it(name, () => {
+				const res = testParser(parser, content)
+				snapshot(res)
+				assert.deepEqual(res.errors, expectedErrors)
+			})
 		}
 	})
 })
--- a/packages/core/test/parser/util.spec.ts
+++ b/packages/core/test/parser/util.spec.ts
@@ -44,92 +44,78 @@ function literal(
 	}
 }
 
-describe('core util parsers', () => {
-	describe('any()', () => {
-		const suites: {
-			content: string
-			parsers: [Parser<AstNode>, ...Parser<AstNode>[]]
-			parserToString: string
-		}[] = [
-			{
-				parsers: [literal('foo'), literal('bar')],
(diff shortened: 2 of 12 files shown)


MulverineX (approved):
Make sure to rebase but besides that, this looks great :+1:

packages/json/src/parser/array.ts
```diff
@@ -1,11 +1,11 @@
 import * as core from '@spyglassmc/core'
 import type { JsonArrayNode } from '../node/index.js'
-import { entry } from './entry.js'
+import { json } from './entry.js'
 
 export const array: core.Parser<JsonArrayNode> = (src, ctx) => {
 	const parser = core.list({
 		start: '[',
-		value: entry,
+		value: json(),
```
SPGoding, line 8: Afaik this will create a new function every time the `array` parser is used to parse something. Probably not a big deal but might be better to store a constant like `entry` in `./entry.ts` with `json(false)`.

TheAfroOfDoom: done in [`5bffcff` (#1132)](https://github.com/SpyglassMC/Spyglass/pull/1132/commits/5bffcffc5f719f60516d5b07be139491b97a9ea7)

packages/json/src/parser/entry.ts
```diff
@@ -44,4 +44,5 @@ export function json(dumpErrors = false): core.Parser<JsonNode> {
 	}
 }
 
-export const entry: core.Parser<JsonNode> = json(true)
+export const entry = jsonParser(true)
```
TheAfroOfDoom, line 47: i removed the explicit typing here since it was already receiving the same type from `json(true)` anyway

packages/json/src/parser/entry.ts
```diff
@@ -20,7 +20,7 @@ const LegalNumberStart = new Set([
 	'9',
 	'-',
 ])
-export function json(dumpErrors = false): core.Parser<JsonNode> {
+export function jsonParser(dumpErrors = false): core.Parser<JsonNode> {
```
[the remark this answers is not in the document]
TheAfroOfDoom: added some doc comments in [`41b5c8b` (#1132)](https://github.com/SpyglassMC/Spyglass/pull/1132/commits/41b5c8be0b4e4a22ecd6b7bce1c71bae3be7ab4d)

take a `peek()`, lmk if you want the wording adjusted

SPGoding (approved):

misode (approved):
I made some changes in order to make the json `entry` parser not take a boolean. Instead the error dumping is now handled by a separate wrapper, to make the intent more clear in the JSON initialize function and also so we don't accidentally dump errors in the future.

packages/core/src/parser/util.ts
```diff
@@ -560,3 +560,14 @@ export function acceptIf<P extends Parser<AstNode>>(
...
+/**
+ * @returns A parser that dumps any parser errors after it finishes parsing.
+ */
+export function dumpErrors<P extends Parser<AstNode>>(parser: P): P {
+	return ((src: Source, ctx: ParserContext) => {
+		const ans = parser(src, ctx)
+		ctx.err.dump()
+		return ans
+	}) as ReturnType<typeof dumpErrors<P>>
+}
```
TheAfroOfDoom, line 573: this makes sense and LGTM. i'll add a commit to this PR with unit tests for this new util parser tonight

TheAfroOfDoom: done in [`50cfd67` (#1132)](https://github.com/SpyglassMC/Spyglass/pull/1132/commits/50cfd676d186581d95b9447c8757e295a4fc38df)

packages/core/test/parser/util.spec.ts
```diff
@@ -43,54 +44,92 @@ function literal(
...
+				parser: dumpErrors(boolean),
+				content: 'bar',
+				expectedErrors: [],
+			},
+		]
+		for (const { name, content, parser, expectedErrors } of suites) {
+			it(name, () => {
+				const res = testParser(parser, content)
+				snapshot(res)
+				assert.deepEqual(res.errors, expectedErrors)
```
misode, line 131: If you're snapshotting the result, there is no reason to check for the expectedErrors again.

misode: Additionally, this file is already called `core/test/parser/util`, so it's unnecessary to wrap it in `describe('core util parsers'`

TheAfroOfDoom: > If you're snapshotting the result, there is no reason to check for the expectedErrors again.

i kinda like the redundancy in the check -- snapshots are definitely nice to get a big picture of what the parser outputs! but i like also checking for the main thing we want the parser to be doing (e.g. do stuff around errors in this case)

i'm pretty sure we actually have other tests that intermingle `snapshot` and `assert`, so this wouldnt be the first time we do that redundantly:

https://github.com/SpyglassMC/Spyglass/blob/b66297d1f6c6a7bfa34ad2983f55a5e1cba8a278/packages/core/test/common/StateProxy.spec.ts#L88-L110

misode: I was mostly basing this on the above `any()` tests in the same file. I'm not sure how the mentioned test supports your argument. It snapshots the `proxy` object, but then asserts the `testObj` and `barSymbol`. The `expectedErrors` you wrote in the test are exactly the same as the ones in the snapshot, so they're useless.

PR #1133, opened 2024-05-13
🐛 Allow doc comments on mcdoc dispatch statements

misode:
It came up that it would be really useful to allow doc comments in front of dispatch statements, which would simply target the dispatched type.

## Todo
- [x] Update the parser and nodes
- [ ] Update the spec


MulverineX:
A more comprehensive PR is on the way, ideally fixing all broken doc comments

misode:
➡️ #1135

PR #1135, opened 2024-05-13
🐛 Fix doc comment parsing in mcdoc

MulverineX:
Appears to be due to `{ get: () => type }` greedily capturing all nodes after `dispatch` and `type` statements, results in all comments after these to be parsed as dev comments and inserted into the value (`type/dispatcher` or `type/reference`) children.

**Not ready to merge**


misode:
@MulverineX I don't think PR's should be part of the milestone

PR #1136, opened 2024-05-13
🐛 Fix how locales with subtags are handled

### LINKED ISSUE #1134: Make `loadLocale()` fall back gracefully when locale has subtags
Current Behavior:

* Call `loadLocale('en-us')`
* It throws an exception because there is no language file `en-us.js`

Expected Behavior:

* Call `loadLocale('en-us')`
* It should try to load `en-us.js` and fail
* It should try to load `en.js` (with `-us` subtag removed) and succeed

SPGoding:
Fixes #1134.

### DIFF
--- a/packages/locales/src/index.ts
+++ b/packages/locales/src/index.ts
@@ -37,11 +37,33 @@ export function localeQuote(content: string) {
 }
 
 /**
- * @param locale An [IETF language tag](https://en.wikipedia.org/wiki/IETF_language_tag). Defaults to `en`.
+ * @param locale A case-insensitive language tag following the format of
+ * `\w+(-\w+)*` that should ideally correspond to the file name of one of the
+ * files under `locales/`. Defaults to `en`.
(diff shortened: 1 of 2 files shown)


packages/locales/src/index.ts
```diff
@@ -37,11 +37,33 @@ export function localeQuote(content: string) {
...
+		// Most likely due to unknown locale.
+
+		const lastDashIndex = locale.lastIndexOf('-')
+		if (lastDashIndex === -1) {
+			// The locale has no subtags. No more locales to try.
+			throw e
+		}
+
+		// Try again with the rightmost subtag removed.
+		return loadLocale(locale.slice(0, lastDashIndex), dry)
```
misode, line 66: This makes me a bit nervous but I can't see any way this could get in an infinite loop, so it should be fine.

misode (approved):

TheAfroOfDoom (approved):

PR #1138, opened 2024-05-14
♻️ Use LSP to provide inlay hints

### LINKED ISSUE #1137: Use LSP implementation of inlay hints
Inlay hint was not standarized in LSP 2 years ago so it was implemented with a custom `spyglassmc/inlayHints` request:

https://github.com/SpyglassMC/Spyglass/blob/50b9c24d3a4500483dcf1ac23bc1af65a2e8bcda/packages/vscode-extension/src/extension.mts#L69-L108

We should be able to remove this entire chunk of code and implement it properly in `language-server/src/server.ts` now. I call this tactical two-year delay a massive success /s

SPGoding:
Resolves #1137.

* Updates dependencies on `vscode-languageserver` and `vscode-languageclient`.
* Uses LSP to provide inlay hints instead of our own protocol.
* Enables inlay hints on more command nodes by default:
  * 'boolean',
  * 'double',
  * 'float',
  * 'integer',
  * 'long',
  * 'mcfunction:coordinate',
  * 'mcfunction:vector',
  * 'mcfunction:command_child/unknown'
  * The rationale behind these nodes is that they're fairly "simple" (subjectively) that it may not be immediately clear to an inexperienced command user what the argument is for. e.g. the boolean argument in `effect give`, the various vectors in `clone`, `tp`, etc, while a more "complicated" node like a block ID is more obvious that it represents a type of block.
    inlay hints

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -34,39 +34,6 @@
 		},
 		{
 			"type": "npm",
-			"script": "watch",
-			"group": "build",
-			"isBackground": true,
-			"path": "packages/vscode-extension/",
-			"dependsOn": "npm: watch",
-			"problemMatcher": {
-				"severity": "error",
-				"fileLocation": "absolute",
-				"source": "esbuild",
-				"background": {
-					"activeOnStart": true,
-					"beginsPattern": "Start building\\.\\.\\.",
-					"endsPattern": "Built successfully\\."
-				},
-				"pattern": [
-					{
-						"regexp": "ERROR in ([^\\(]*)\\((\\d+),(\\d+)\\):",
-						"file": 1,
-						"line": 2,
-						"column": 3
-					},
-					{
-						"regexp": "([A-Za-z0-9-]+):(.*)",
-						"message": 2,
-						"code": 1
-					}
-				]
-			},
-			"label": "npm: watch - packages/vscode-extension",
-			"detail": "Builds the VS Code extension."
-		},
-		{
-			"type": "npm",
 			"script": "test",
 			"group": "test",
 			"problemMatcher": [],
--- a/packages/language-server/src/server.ts
+++ b/packages/language-server/src/server.ts
@@ -12,8 +12,6 @@ import type {
 	CustomInitializationOptions,
 	CustomServerCapabilities,
 	MyLspDataHackPubifyRequestParams,
-	MyLspInlayHint,
-	MyLspInlayHintRequestParams,
 } from './util/index.js'
 import { toCore, toLS } from './util/index.js'
 
@@ -112,12 +110,16 @@ connection.onInitialize(async (params) => {
 			},
 		})
 		service.project
-			.on('documentErrored', ({ errors, uri, version }) => {
-				connection.sendDiagnostics({
-					diagnostics: toLS.diagnostics(errors),
-					uri: uri,
-					version: version,
-				})
+			.on('documentErrored', async ({ errors, uri, version }) => {
+				try {
+					await connection.sendDiagnostics({
+						diagnostics: toLS.diagnostics(errors),
+						uri: uri,
+						version: version,
+					})
+				} catch (e) {
+					console.error('[sendDiagnostics]', e)
+				}
 			})
 			.on('ready', () => {
 				progressReporter?.done()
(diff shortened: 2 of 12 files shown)


.vscode/tasks.json
MulverineX, line 1: Is this superseded by something else?

MulverineX (changes requested):

.vscode/tasks.json
SPGoding: See commit message in e91ae35bcd5103f0c14f8835e9df474f83cb0477

MulverineX (approved):

packages/language-server/src/server.ts
```diff
@@ -112,17 +110,16 @@ connection.onInitialize(async (params) => {
...
-			.on('documentErrored', ({ errors, uri, version }) => {
-				if (!core.fileUtil.isFileUri(uri)) {
-					// Don't send errors for non-file URIs to the client.
-					return
+			.on('documentErrored', async ({ errors, uri, version }) => {
+				try {
+					await connection.sendDiagnostics({
+						diagnostics: toLS.diagnostics(errors),
+						uri: uri,
+						version: version,
```
TheAfroOfDoom, line 118: i know you just moved this object and it's not actually new, but it made me think of [`object-shorthand`](https://eslint.org/docs/latest/rules/object-shorthand)

definitely more of a style-focused rule though. thoughts?

packages/language-server/src/util/toLS.ts
```diff
@@ -196,17 +195,19 @@ export function hover(hover: core.Hover, doc: TextDocument): ls.Hover {
...
 	hint: core.InlayHint,
 	doc: TextDocument,
-): MyLspInlayHint {
+): ls.InlayHint {
 	return {
 		position: doc.positionAt(hint.offset),
-		text: hint.text,
+		label: hint.label,
+		...hint.paddingLeft ? { paddingLeft: hint.paddingLeft } : {},
+		...hint.paddingRight ? { paddingRight: hint.paddingRight } : {},
```
TheAfroOfDoom, line 203: is this any different than just returning the `paddingLeft` (or right) property directly?

```suggestion
		paddingLeft: hint.paddingLeft,
		paddingRight: hint.paddingRight,
```

as it is now with the spread property is kinda hard to read in comparison, assuming they're the same functionally (which I think they are?)

TheAfroOfDoom (commented):
can you share docs for where you found out about the new version of language server supporting inlay hints natively? i'm struggling to find reading on it. i guess the types when i F12 on `ls.InlayHint` are giving some context at least

---

also left some misc comments/questions

.vscode/tasks.json
```diff
@@ -32,39 +32,6 @@
 			"label": "npm: watch",
 			"detail": "Compiles all packages to JavaScript."
 		},
-		{
-			"type": "npm",
-			"script": "watch",
-			"group": "build",
```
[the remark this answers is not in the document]
SPGoding: there used to be two separate `watch` scripts but I used `wireit` to condense them at one point and forgot to change the VS Code tasks lol.

packages/language-server/src/server.ts
SPGoding: yeah good call. i guess it's not really necessary to enable the lint rule since it's pretty minor but i do prefer the shorthand style as well.

packages/language-server/src/util/toLS.ts
SPGoding: probably the same functionality. i just didn't want to have extra keys whose value is `undefined` (e.g. `{ label: 'hint', paddingLeft: undefined, paddingRight: true }` v.s. `{ label: 'hint', paddingRight: true }`).

SPGoding:
> can you share docs for where you found out about the new version of language server supporting inlay hints natively?

I refer to https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/ and it says it in the `What’s new in 3.17` section

packages/language-server/src/util/toLS.ts
TheAfroOfDoom: actually I'm pretty sure an `undefined` property won't actually show up in an object.

i.e.

```ts
const x = { foo: undefined, bar: 2 }
console.log(x)
// { bar: 2 }
```

PR #1145, opened 2024-05-14
🔨 Better type for numeric mcdoc literals

NeunEinser:
This makes numeric literal types more expressive in their type.

This makes checking these types against nodes a bit less cumbersome in my current related work.


packages/mcdoc/src/binder/index.ts
```diff
@@ -1230,11 +1229,16 @@ function convertLiteralValue(
 function convertLiteralNumberSuffix(
 	node: LiteralNode | undefined,
 	ctx: McdocBinderContext,
-): LiteralNumberSuffix | undefined {
+): (typeof NumericTypeKinds)[number] | undefined {
```
SPGoding, line 1233: Can we use `NumericTypeIntKind` instead of `(typeof NumericTypeIntKinds)[number]`?

NeunEinser:
Looks like there is a failing test now, because of the node type change. I'll wait until I have feedback if this is a desirable change until looking into how to fix those.

packages/mcdoc/src/binder/index.ts
```diff
@@ -1215,9 +1215,9 @@ function convertLiteralValue(
 	} else if (TypedNumberNode.is(node)) {
 		const { suffix, value } = TypedNumberNode.destruct(node)
 		return {
-			kind: 'number',
+			kind: convertLiteralNumberSuffix(suffix, ctx) ??
+				(Number.isInteger(value.value) ? 'int' : 'double'),
```
SPGoding, line 1219: `Number.isInteger(1.0)` would be `true` so we would have an `int` literal number when it is in fact a floating point number. Probably have to change the spec and the parser to properly fix this (the spec current specifies `float` + suffix instead of using `float` and `interger` properly cuz i was too lazy).

SPGoding:
Otherwise I think this would be a good change.

packages/mcdoc/src/binder/index.ts
NeunEinser: Yep, that's less cursed for sure

packages/mcdoc/src/binder/index.ts
NeunEinser: Hmm, good point

NeunEinser: I changed the parser to emit an integer node when there is no dot and no float type marker

NeunEinser:
I can't figure out how to fix the tests. From the output it looks like it tries to see if the current mcdoc exports are equivalent to what this branch would generate, but there is ofc an intentional mismatch. Not sure how to change the test data.

SPGoding:
You can set the environment variable `SNAPSHOT_UPDATE=1` when running `npm test` or just delete the `__snapshots__` folder and let `npm test` re-generate that. We should probably add this to README

NeunEinser:
It deleted `__snapshots__/packages/core/test-out/service/Operations.spec.js`, not sure why.

SPGoding:
> It deleted `__snapshots__/packages/core/test-out/service/Operations.spec.js`, not sure why.

the source file was moved from `service/` to `common/` so that's just an orphan snapshot.

SPGoding (approved):
LGTM, would be nice if you can update the syntax for `TYPED_NUMBER` in the mcdoc spec (`docs/user/mcdoc/index.adoc`) as well. Something like this probably works but it's been a while since I've worked on that punctuation soup document:

```adoc
<<t-typed-number>>::
	<<t-float>> ++[++`d` `D` `f` `F`++]++^?^ | +
	<<t-integer>> ++[++`b` `B` `l` `L` `s` `S`++]++^?^
```

NeunEinser:
(rebase, no new commits)

TheAfroOfDoom:
requested myself as a non-blocking reviewer so i remember to look at this eventually

packages/mcdoc/src/binder/index.ts
```diff
@@ -77,19 +77,19 @@ import {
...
 	DynamicIndex,
 	EnumTypeField,
 	Index,
-	LiteralNumberCaseInsensitiveSuffix,
-	LiteralNumberSuffix,
 	LiteralValue,
 	McdocType,
 	NumericRange,
 	NumericTypeKind,
+	NumericTypeKinds,
```
misode, line 92: Unused import

misode (approved):

__snapshots__/packages/core/test-out/service/StateProxy.spec.js
TheAfroOfDoom, line 1: i might prefer we move these file deletions into a separate PR that removes the orphaned snapshots? as-is in here it's almost implying that these mcdoc type changes caused these file diffs

TheAfroOfDoom: maybe more leg-work than is worth though

packages/mcdoc/src/type/index.ts
```diff
@@ -136,45 +136,25 @@ export interface StringType {
...
+	value: boolean
+}
+export interface LiteralStringValue {
+	kind: 'string'
+	value: string
+}
+export interface LiteralNumericValue {
+	kind: NumericTypeKind
+	value: number
+}
```
TheAfroOfDoom, line 153: this is really nice

__snapshots__/packages/mcdoc/test-out/__fixture__/struct/simple.spec.js
```diff
@@ -47,7 +47,7 @@ exports['mcdoc __fixture__ struct/simple 1'] = {
                     "value": {
                       "kind": "literal",
                       "value": {
-                        "kind": "number",
+                        "kind": "double",
                         "value": 1.19
```
TheAfroOfDoom, line 51: is there any meaningful difference in this diff? does it matter that we parse release versions as a "double" rather than just some number?

we should probably parse it as a "semantic version" type instead, but that's prob better for a later PR

TheAfroOfDoom (commented):
couple smaller things. generally this seems like a good idea though, nice work 😸 

---

@SPGoding what are your thoughts on removing `/__snapshots__/** linguist-generated=true` from `.gitattributes`?

i guess they technically _are_ generated, but it'd be nice if they weren't auto not rendered in the `Files changed` view. i find myself definitely wanting to review each snapshot change for robustness

---

similarly, why is `README.md` set as `linguist-generated=true`?

__snapshots__/packages/mcdoc/test-out/__fixture__/struct/simple.spec.js
NeunEinser: It's parsed as string when it's `1.19.3`.

According to mcdoc spec, when not specifying a number suffix, it is parsed as either an int or a double, depending on its value.

__snapshots__/packages/core/test-out/service/StateProxy.spec.js
NeunEinser: I could revert the changes to these deleted files, just not sure if it's wanted.

SPGoding:
> what are your thoughts on removing /__snapshots__/** linguist-generated=true from .gitattributes?

i agree

> similarly, why is README.md set as linguist-generated=true?

it used to have automatically generated contributor information. i guess we can remove this as well.

__snapshots__/packages/core/test-out/service/StateProxy.spec.js
TheAfroOfDoom: eh its probably fine to keep as-is. would be nice if we eventually had a script to check for orphaned snapshots in GHA

TheAfroOfDoom (approved):

PR #1149, opened 2024-05-14
🐛 Improve version and release target resolution

misode:
Instead of using a map of regexes to convert the pack format to a release version, it loops over all versions to find the correct release version.
- If the pack format is higher than the latest release, use the latest snapshot
- Look for versions from recent to oldest, picking the most recent release that matches
- If the pack format is still lower than the oldest version, simply use the oldest release (which is 1.14)

Additionally, this PR fixes the issue when the `release_target` field is missing in versions after 22w42a
- If the version is already a release, us its own `id` field
- It tries to find the next release cronologically
- If there is no release, hardcode the current release target (this may be improved in the future)

### DIFF
--- a/packages/java-edition/src/dependency/mcmeta.ts
+++ b/packages/java-edition/src/dependency/mcmeta.ts
@@ -1,6 +1,5 @@
 import * as core from '@spyglassmc/core'
 import type { PackMcmeta, ReleaseVersion, VersionInfo } from './common.js'
-import { PackVersionMap } from './common.js'
 
 /**
  * @param inputVersion {@link core.Config.env.gameVersion}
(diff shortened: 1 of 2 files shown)


packages/java-edition/src/dependency/mcmeta.ts
```diff
@@ -34,23 +47,39 @@ export function resolveConfiguredVersion(
...
 		}
-		return toVersionInfo(0)
+		// Fall back to the latest release if pack mcmeta is not available
+		return toVersionInfo(latestRelease)
 	} else if (inputVersion === 'latest release') {
-		return toVersionInfo(versions.findIndex((v) => v.type === 'release'))
+		return toVersionInfo(latestRelease)
 	} else if (inputVersion === 'latest snapshot') {
-		return toVersionInfo(versions.findIndex((v) => v.type === 'snapshot'))
+		return toVersionInfo(versions[0])
```
SPGoding, line 79: good catch!

SPGoding (approved):

PR #1150, opened 2024-05-15
🐛 Allow single quotes in NBT paths

### LINKED ISSUE #1058: Allow single quotes in NBT paths
https://bugs.mojang.com/browse/MC-175504

SPGoding:
Fixes #1058. This change allows single quotes for
all versions, not just the versions after 1.20 Pre- release 2, as the `nbt` package does not have
access to the `ReleaseVersion.cmp()` function from `java-edition` and it'd take too much effort to
do it properly just to eliminate a tiny little
false negative. We may revisit this if we decide
to use pack formats internally instead since it'd
be an easy
`parseInt(ctx.project['loadedFormat']) < 15` check.

### DIFF
--- a/__snapshots__/packages/nbt/test-out/parser/path.spec.js
+++ b/__snapshots__/packages/nbt/test-out/parser/path.spec.js
@@ -82,16 +82,7 @@ exports['nbt path() Parse "\'foo\'" 1'] = {
       "end": 5
     }
   },
-  "errors": [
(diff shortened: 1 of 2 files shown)


TheAfroOfDoom (approved):
looks like snapshots need to be updated but LGTM otherwise

NeunEinser (approved):

PR #1155, opened 2024-05-17
🐛 Fix message reporting breaking on empty string

NeunEinser:
When sending an empty string to error reporting, no errors show up in the entire file.

This fixes this in two ways:
- When trying to send an empty string to b reported, it's just ignored and no error is reported for that line
- When trying to localize an unknown key, instead of returning an empty string, the key itself is returned
   - This could have side effects in other code that relies on a broken key to return nothing. That seems like a potential bug though, so I think highlighting that with showing the translation key is not bad.
   - Potentially, we could also have it depend on some sort of 'dev' flag which to not show keys in a release build
   - When developing, this illustrates the kind of mistake you made better. Instead of not showing an error, which may lead you thinking your error reporting logic is somehow wrong, you immediately see that you are missing a translation key


packages/core/src/service/ErrorReporter.ts
```diff
@@ -16,9 +16,11 @@ export class ErrorReporter {
 		severity = ErrorSeverity.Error,
 		info?: LanguageErrorInfo,
 	): void {
-		this.errors.push(
-			LanguageError.create(message, Range.get(range), severity, info),
-		)
+		if (message.trim() !== '') {
```
SPGoding, line 19: Is there a case where we intentionally create an error with an empty message that should be ignored? I feel like if an error with empty message is created it's most certainly caused by a bug somewhere and we shouldn't just silently ignore them.

NeunEinser: I could throw an exception too, if you'd prefer that.

SPGoding: That would probably be better (as long as we catch all of them during development lol)

NeunEinser: ~~Just wrap all instances of err.report into a try catch~~

NeunEinser:
Looks like there are some missing translation keys for longs, as illustrated by the test changes. I can add them in this PR, I think it would make sense?

misode (approved):

PR #1158, opened 2024-05-17
✨ Improve JSON number parsing result

NeunEinser:
This examines the number at parse time to determine whether it is an integer or a float. This allows for better checking later on, when someone tries to define decimal places for a number that is defined as an integer.

Additionally, I simplified some of the other JSON parser definitions to use `core.setType` instead of doing that manually.


packages/json/src/formatter/index.ts
```diff
@@ -44,7 +44,6 @@ export function register(meta: MetaRegistry): void {
 		core.formatter.boolean,
 	)
 	meta.registerFormatter<JsonNullNode>('json:null', () => 'null')
-	meta.registerFormatter<JsonNumberNode>('json:number', core.formatter.float)
```
NeunEinser, line 47: I am pretty sure this should not be needed anymore, as the json number node now has a child of either long or float, which should be formatted by the core formatters already.

misode: "pretty sure"? Did you test this?

NeunEinser: no, the formatter only returns a toString for float, so I don't know if there is any observable change when it's therers when it's not there.

```ts
export const float: Formatter<FloatBaseNode> = (node) => {
	return node.value.toString()
}
```

NeunEinser: Ah, actually there is an issue here, it complains about not finding the json:number formatter.

packages/json/src/node/JsonAstNode.ts
```diff
@@ -156,9 +156,10 @@ export namespace JsonStringNode {
 export interface JsonNumberExpectation extends JsonBaseExpectation {
 	readonly type: 'json:number'
 }
-export interface JsonNumberNode extends core.FloatBaseNode, JsonBaseAstNode {
+export interface JsonNumberNode extends JsonBaseAstNode, core.AstNode {
 	readonly type: 'json:number'
-	readonly value: number
+	readonly children: [core.LongNode | core.FloatNode]
```
misode, line 161: Does this need to be a `LongNode`? Wouldn't `IntegerNode` be enough, because this is JSON?

packages/json/test/parser/number.spec.ts
```diff
@@ -0,0 +1,30 @@
+import type { LanguageError } from '@spyglassmc/core'
+import {
+	showWhitespaceGlyph,
+	testParser,
+} from '@spyglassmc/core/test-out/utils.js'
+import assert from 'assert'
```
misode, line 6: `assert` an `LanguageError`: unused imports

packages/json/src/node/JsonAstNode.ts
NeunEinser: I used long node because it uses a bigint internally, so the parsed value can be more precise.

misode: That's what I mean. Does it need that precision? Aren't JSON values inherently restricted by the precision of JS numbers?

NeunEinser: I just figured there isn't any disadvantage of using a bigint. Not sure what the json spec says, and also not sure how Mojang handles this when codec specifies a long.

misode: Fair. I think bigints are a bit annoying in JS, for example if you have `int @ 1..2` and you parse the `1..2` range as numbers but the json number itself is a bigint, you need to convert one to the other to compare them. Either way no major issues so I'm fine with it.

misode (approved):

SPGoding (approved):

PR #1177, opened 2024-05-19
🐛 Fix computing relative URIs

SPGoding:


packages/core/src/service/fileUtil.ts
```diff
@@ -25,7 +32,7 @@ export namespace fileUtil {
 		rootUris: readonly RootUriString[],
 	): Generator<string, undefined, unknown> {
 		for (const root of rootUris) {
-			if (uri.startsWith(root)) {
+			if (uriStartsWith(uri, root)) {
 				yield decodeURIComponent(uri.slice(root.length))
```
misode, line 36: This still causes problems, because the `uri.slice(root.length)` is using the length of an encoded uri to slice a decoded uri, which causes the modules to look like `va/server/...`

misode (changes requested):

packages/core/src/service/fileUtil.ts
SPGoding: great point. urghh.

misode (approved):
Seems to work as expected. Maybe a tad overcomplicated to my taste.

Would something like this not have worked?
```ts
const decodedUri = decodeURIComponent(uri)
for (const root of rootUris) {
  const decodedRoot = decodeURIComponent(root)
  if (decodedUri.startsWith(decodedRoot)) {
    yield decodedUri.slice(decodedRoot.length)
  }
}
```

SPGoding:
> Would something like this not have worked?
> 
> ```ts
> const decodedUri = decodeURIComponent(uri)
> for (const root of rootUris) {
>   const decodedRoot = decodeURIComponent(root)
>   if (decodedUri.startsWith(decodedRoot)) {
>     yield decodedUri.slice(decodedRoot.length)
>   }
> }
> ```

It probably will work for our use case, but if there is some weird encoded special characters in a path segment it won't work. e.g. `getRels('file:///foo%2Fbar/qux', ['file:///foo/'])` (`%2F` is URL encoded `/`) will return `['bar/qux']` even though `foo%2Fbar/` is the actual first name and is not under the root `foo/`.

packages/core/src/service/fileUtil.ts
```diff
@@ -6,6 +6,45 @@ export type RootUriString = `${string}/`
...
+		}
+
+		const baseComponents = baseUri.pathname.split('/')
+			.filter((v) => !!v)
+		const targetComponents = targetUri.pathname.split('/')
+			.filter((v) => !!v)
+
+		for (const [i, c] of baseComponents.entries()) {
+			if (
+				decodeURIComponent(c) !== decodeURIComponent(targetComponents[i])
```
misode, line 33: I think this will throw if `targetComponents` happens to be shorter than `baseComponents`

SPGoding: It doesn't throw but it's weird and won't work if base is `file:///foo/undefined/` and target is `file:///foo/`.

image

MulverineX (approved):

PR #1178, opened 2024-05-19
🐛 `/mcdoc-cli`: Remove `java-edition`, add cwd, fix shebang, revert m…

MulverineX:
…ts nonsense, bump version


packages/mcdoc-cli/src/index.ts
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env -S tsx
+#!/usr/bin/env -S node
```
SPGoding, line 1: probably doesn't need `-S` anymore (i don't think it hurts to keep it though)

SPGoding (approved):

PR #1179, opened 2024-05-19
Implement 1.20.5+ item slots

### LINKED ISSUE #1159: Update item slots
Some new slots are not available in the current `item_slot` parser (for example `contents` or `armor.body`). Special care should be taken that these new slot values are only available in the new versions by comparing the currently loaded version.

This could be solved together with #1157, but if it isn't; check that the new item slots are also availble for this `item_slots` parser.

Nico314159:
Resolves #1159


MulverineX (changes requested):
These additions need to use `ReleaseVersion.cmp`

Nico314159:
Okay, I think I made the whole pipeline work with `ctx`, but for some reason all the tests are failing...

Nico314159:
@MulverineX Take a look when you get a chance please

package.json
```diff
@@ -72,7 +72,7 @@
 		"source-map-support": "^0.5.21",
 		"ts-loader": "^9.4.2",
 		"ts-node": "^10.9.1",
-		"typescript": "^5.0.4",
+		"typescript": "^5.4.5",
```
misode, line 75: Why was this change made in this PR? (also another change in this file)

Nico314159: I didn't do it on purpose, I think npm did it automatically when I ran `npm i`.

packages/java-edition/src/mcfunction/parser/argument.ts
```diff
@@ -87,6 +87,7 @@ function shouldValidateLength(ctx: core.ParserContext) {
  */
 export const argument: mcf.ArgumentParserGetter = (
 	rawTreeNode,
+	ctx: core.ContextBase
 ): core.Parser | undefined => {
```
misode, line 91: It makes sense to pass the `ctx` to the completer `getMockNodes`, but this function is already returning a `Parser`, which is a function that has access to `ctx`.

packages/java-edition/src/mcfunction/parser/argument.ts
```diff
@@ -189,7 +190,7 @@ export const argument: mcf.ArgumentParserGetter = (
 		case 'minecraft:item_predicate':
 			return wrap(itemPredicate)
 		case 'minecraft:item_slot':
-			return wrap(core.literal(...ItemSlotArgumentValues))
+			return wrap(core.literal(...getItemSlotArgumentValues(ctx)))
```
misode, line 193: This is how you could get the `ctx` for the item slots without requiring the entire `argument` function to need the context.
```suggestion
			return wrap((src, ctx) => {
			  return core.literal(...getItemSlotArgumentValues(ctx))(src, ctx)
			})
```

packages/java-edition/src/mcfunction/common/index.ts
```diff
@@ -11,24 +12,34 @@ export const GamemodeArgumentValues = [
...
+		'weapon.mainhand',
+		'weapon.offhand',
+	]
+	if (ReleaseVersion.cmp(release, '1.20.5') >= 0) {
+		output.push(...[...Array(4).keys()].map((n) => `player.crafting.${n}`),
+		            'armor.body', 
+		            'contents', 
+		            'player.cursor')
+	}
+	else { output.push('horse.armor') }
```
misode, line 40: The formatting of this is a bit off

misode (changes requested):

packages/java-edition/src/mcfunction/parser/argument.ts
Nico314159: I'm having trouble with grokking this intuitively.

SPGoding: `core.literal()` is a function that returns a `Parser` (a `Parser` is (basically) a function in the form of `(src, ctx) -> AstNode`). It is functionally equivalent to write:

```ts
core.literal(...)
```

and 

```ts
(src, ctx) => core.literal(...)(src, ctx)
```

as the first one returns a `Parser` directly, and in the second one we are writing a parser (the `(src, ctx) => ...` part) that returns the result of calling the `core.literal(...)` parser with `(src, ctx)`. Misode's suggestion turns the original form into the second form so `getItemSlotArgumentValues()` has access to `ctx` from the `Parser` being called.

MulverineX (approved):

PR #1180, opened 2024-05-19
✨ Add parser support for backslash continuation

### LINKED ISSUE #1069: Support backslash continuation in functions

TheAfroOfDoom:
# Summary

closes #1069 

this PR adds backslash continuation for `command` and `argument` parsers in mcfunctions.

## Preview

| `command` parser | `argument` parser _(+ node range highlighting correctly)_ |
|-|-|
| image | image |
| `command` and `argument` parser together | correct index mappings across lines |
| image | image |
| `command` and `argument` completers not working yet | ...but completers do work as long as you're not splitting the node (?) up between lines? |
| completer not working summon | completer sorta working entity selectors |

## Next steps

_(these should be ticketized)_

- fix colorizer not really working across lines (e.g. open/close brackets `[`/`]` across lines)
  - plenty of examples of this in the above images
- fix command completer mostly not working


SPGoding:
amazing work! i'll look over this when i get more time.

> open/close brackets [/] across lines

i believe that is not an us issue. we don't colorize brackets. do you have some other extensions?

packages/core/src/parser/util.ts
```diff
@@ -506,6 +506,71 @@ export function stopBefore<N extends Returnable>(
...
+			indexMap.push({
+				inner: Range.create(wrappedStr.length),
+				outer: Range.span(from, to),
+			})
+		}
+
+		const wrappedSrc = new Source(
+			wrappedStr,
+			IndexMap.merge(src.indexMap, indexMap),
+		)
```
NeunEinser, line 566: I suppose Index Map is sth that when you report a range it maps it accordingly? That's neat!

TheAfroOfDoom: yeah this took me a bit to understand. basically it's just saying "here's what the parser will see, but here's what character ranges that maps to in vscode's actual text document"

SPGoding (commented):
the completion is computed correctly but VS Code doesn't display them.

completion computed in log

i think this happens because VS Code uses text in the `replace` range to filter through the options, which does not match any of the `label` of the completion items. e.g., the `replace` range is `[(8, 0); (9, 3))`, so the source text VS Code uses to filter options is `ad\<LF>huh`. the `label` of the completion item is `advancement`, which does not match the source text, so no completion item is shown. one way i can think of to fix this is to add `filterText` like `ad\<LF>vancement` to the completion items so VS Code can filter the options properly (and if we can construct those `filterText` might as well set the `newText` to them so the user gets to keep their insane word-breaking line continuity)

packages/core/src/parser/util.ts
```diff
@@ -506,6 +506,71 @@ export function stopBefore<N extends Returnable>(
...
+				wrappedStr += src.read()
+				continue
+			}
+
+			// Create an index map that skips the trailing backslash + whitespace
+			const from = src.getCharRange()
+			// skip the `\`
+			src.skip()
+			// skip trailing whitespace + newline + preceding whitespace on the next line
+			src.skipWhitespace()
```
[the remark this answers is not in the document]
TheAfroOfDoom: fixed in [`059a825` (#1180)](https://github.com/SpyglassMC/Spyglass/pull/1180/commits/059a825c4ea4e799b9ed2ebb17929e492d9a2b33)

SPGoding:
Would it make sense to wrap `concatOnTrailingBackslash` on the entire `command` parser, instead of on individual argument parsers? With the current implementation the built-in parsers in `mcfunction` package (trailing content and unknown parser) does not support backslash continuation.

For something like this,

```mcfunction
advancement ! \
aartsarst
```

The `!` is unparsable, so anything after it is considered to be trailing content. The expected trailing content should expand into the second line, but with the current implementation the second line is treated as the start of a new command.

image

TheAfroOfDoom:
> Would it make sense to wrap `concatOnTrailingBackslash` on the entire `command` parser

i believe it already was wrapped around the `command` parser:

https://github.com/TheAfroOfDoom/Spyglass/blob/059a825c4ea4e799b9ed2ebb17929e492d9a2b33/packages/mcfunction/src/parser/command.ts#L96-L103

however...

> The `!` is unparsable, so anything after it is considered to be trailing content. The expected trailing content should expand into the second line, but with the current implementation the second line is treated as the start of a new command.

good catch. when looking into where exactly we should place the `concat` parser i was reminded of our `comment` and `command_macro` parsing that is done in the mcfunction `entry` parser

this made me wonder how Minecraft would parse something like this:

```mcfunction
# a comment \
say i will not say
```

turns out it prints nothing, meaning the backslash continuation is the first thing that happens when Minecraft parses a `.mcfunction` file (prior to even comment parsing). this isn't too crazy but was interesting to me nonetheless

the fix for this is fairly simple and actually lets me simplify some code/remove unnecessary test code. commits soonTM

packages/core/src/parser/util.ts
```diff
@@ -506,6 +506,65 @@ export function stopBefore<N extends Returnable>(
...
+		const wrapper = '\\'
+		const wrappedSrcCursor = wrappedStr.length
+		const indexMap: IndexMap = []
+
+		while (src.canRead()) {
+			wrappedStr += src.readUntil(wrapper)
+			if (!src.canRead()) {
+				// Stop wrapping `src` if we reach end of file or a terminator before a wrapper
+				break
+			}
```
TheAfroOfDoom, line 537: this seems like it could be more concise but i can't think if how immediately

.vscode/launch.json
```diff
@@ -26,6 +28,7 @@
 			"request": "launch",
 			"name": "Run Unit Tests",
 			"program": "${workspaceFolder}/node_modules/mocha/bin/_mocha",
+			"args": ["--timeout", "999999"],
```
misode, line 31: This probably shouldn't be part of this PR?

packages/core/src/parser/util.ts
```diff
@@ -506,6 +506,65 @@ export function stopBefore<N extends Returnable>(
...
+): InfallibleParser<N>
+export function concatOnTrailingBackslash<N extends Returnable>(
+	parser: Parser<N>,
+): Parser<N>
+export function concatOnTrailingBackslash<N extends Returnable>(
+	parser: Parser<N>,
+): Parser<N> {
+	return (src, ctx): Result<N> => {
+		let wrappedStr = src.sliceToCursor(0)
+		const wrapper = '\\'
```
misode, line 528: Is this variable necessary, can't you just inline the `'\\'`? Or alternatively make it a function argument that defaults to `\\`.

packages/core/src/parser/util.ts
```diff
@@ -506,6 +506,65 @@ export function stopBefore<N extends Returnable>(
...
+		const wrappedSrcCursor = wrappedStr.length
+		const indexMap: IndexMap = []
+
+		while (src.canRead()) {
+			wrappedStr += src.readUntil(wrapper)
+			if (!src.canRead()) {
+				// Stop wrapping `src` if we reach end of file or a terminator before a wrapper
```
misode, line 535: What does "or a terminator" mean here? What other option is there except reaching EOF?

packages/mcfunction/test/parser/utils.ts
```diff
@@ -0,0 +1,34 @@
+import type { RootTreeNode } from '../../../java-edition/lib/dependency/mcmeta.js'
```
misode, line 1: The `mcfunction` package also has a `RootTreeNode` type, you should probably use that one.

packages/mcfunction/test/parser/entry.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+		{ content: '# this is a comment \\ \n still a comment' },
+		{ content: '$ this is a macro \\ \n this is still a macro' },
+		{ content: 'sa\\  \n  y \\ \n hi' },
+		{ content: 'say trailing \\\n data' },
+		{ content: 'say \\\n hi \n # comment start \\\n end \n say hi' },
+	]
+	for (const { content } of cases) {
+		it(`Parse "${showWhitespaceGlyph(content)}"`, () => {
+			const release = '1.99'
+			initialize(mockProjectData(), tree, release)
```
misode, line 30: I'm not sure if it's a good idea to use the global state `initialize` function in a test. But I see that fixing that would require refactoring how the mcfunction parser works in a way where you pass it a tree instead of just the tree name.

packages/mcfunction/test/parser/utils.ts
TheAfroOfDoom, line 1: the type of this changed from the `RootTreeNode` in `packages\mcfunction\src\tree\type.ts` to `packages\java-edition\src\dependency\mcmeta.ts`

packages/core/src/parser/util.ts
TheAfroOfDoom: oops outdated comment

packages/mcfunction/test/parser/utils.ts
TheAfroOfDoom: that one leads to type errors.

image

image

packages/mcfunction/test/parser/entry.spec.ts
TheAfroOfDoom: > fixing that would require refactoring how the mcfunction parser works in a way where you pass it a tree instead of just the tree name

yeah exactly. i wanted to minimize setup lines for these tests, this felt reasonable to me. but i'm open to alternatives

packages/mcfunction/test/parser/utils.ts
misode: Okay I guess that is caused by using the `initialize` function from the `java-edition` package in a `mcfunction` test. Which feels wrong, as discussed in the other comment.

TheAfroOfDoom:
still need to fix build

packages/mcfunction/test/parser/entry.spec.ts
TheAfroOfDoom: found a good alternative. it was much easier than i thought it'd be

TheAfroOfDoom: also fixes the build!

.vscode/launch.json
TheAfroOfDoom: ill pull it out

packages/mcfunction/test/parser/entry.spec.ts
misode: Much nicer!

.vscode/launch.json
TheAfroOfDoom: posterity https://github.com/SpyglassMC/Spyglass/pull/1189

packages/mcfunction/src/parser/entry.ts
```diff
@@ -52,3 +52,12 @@ export function entry(
...
 })
+
+/**
+ * Concatenates lines together on trailing backslashes before running the main
+ * {@link mcfunction} parser
+ */
+export const entry = (
+	commandTreeName: string,
+	argument: ArgumentParserGetter,
+) => core.concatOnTrailingBackslash(mcfunction(commandTreeName, argument))
```
SPGoding, line 63: This should only be enabled for a certain version. I think we could either add a new parameter to the `entry` function `supportsBackslashContinuation` to toggles the `core.concatOnTrailingBackslash()` wrap, or move the wrap to `java-edition/mcfunction`.

TheAfroOfDoom: done in [`c16a08a` (#1180)](https://github.com/SpyglassMC/Spyglass/pull/1180/commits/c16a08a191a2d157d981528a4ab7d1efd782ec4a)

TheAfroOfDoom: the flag param is there now, but i think i'd prefer to do the actual version checking -> gating to after `1.20.2` in a separate PR since this one's already pretty big at this point

packages/core/src/parser/util.ts
```diff
@@ -506,6 +506,63 @@ export function stopBefore<N extends Returnable>(
...
+): Parser<N> {
+	return (src, ctx): Result<N> => {
+		let wrappedStr = src.sliceToCursor(0)
+		const wrappedSrcCursor = wrappedStr.length
+		const indexMap: IndexMap = []
+
+		while (src.canRead()) {
+			wrappedStr += src.readUntil('\\')
+			if (!src.canRead()) {
+				break
```
misode, line 534: When this happens, Minecraft throws an error `Line continuation at end of file`, so we should probably show an error in vscode as well.

misode (changes requested):

packages/core/src/parser/util.ts
NeunEinser: What happens when there is a \ at the end of the line followed by an empty line?

TheAfroOfDoom: > When this happens, Minecraft throws an error `Line continuation at end of file`, so we should probably show an error in vscode as well.

confirmed this in-game. implementing fix

---

> What happens when there is a \ at the end of the line followed by an empty line (or an empty line before EOF)?

looks like it's fine as long as its not right before EOF

the following `reload`s fine in-game and outputs as the following:

image

image

TheAfroOfDoom:
apparently i have no idea how to cleanly do a merge commit RIP

packages/core/src/parser/util.ts
TheAfroOfDoom: [`0548d94` (#1180)](https://github.com/SpyglassMC/Spyglass/pull/1180/commits/0548d94fcdf84656a3532de7bd3d8afdde983a8f)

SPGoding (approved):

packages/core/src/parser/util.ts
```diff
@@ -506,6 +507,77 @@ export function stopBefore<N extends Returnable>(
...
+
+			// Minecraft raises a `Line continuation at end of file` if a backslash
+			// (+ optional whitespace to the next line) is right before the end of the file
+			if (!src.canRead()) {
+				const ans: ErrorNode = {
+					type: 'error',
+					range: Range.span(from, src),
+				}
+				ctx.err.report(
+					localize('mcfunction.parser.line-continuation-end-of-file'),
```
misode, line 557: Either move this parser to the `mcfunction` package, make this a more generic core parser message, or allow the function caller to supply a message.

TheAfroOfDoom: [`163283e` (#1180)](https://github.com/SpyglassMC/Spyglass/pull/1180/commits/163283ec04e09f7317fb45c494a1b6672843df57)

misode (approved):

NeunEinser (approved):

MulverineX (approved):

PR #1181, opened 2024-05-19
🔧 Add `import/no-duplicates` lint rule

TheAfroOfDoom:
# Summary

In https://github.com/SpyglassMC/Spyglass/pull/1180 I noticed that my duplicate import statements weren't being caught by the linter. There's (at least) two lint rules that exist that can detect/fix these easily, so this PR's primary purpose is to add one of those lint rules.

ESLint's built-in rule relating to duplicate imports [`no-duplicate-imports`](https://eslint.org/docs/latest/rules/no-duplicate-imports) doesn't have auto-fix for some reason. So instead we went with the [`eslint-plugin-import`](https://github.com/import-js/eslint-plugin-import) plugin which has [`import/no-duplicates`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-duplicates.md) (which can auto-fix these errors).

| sample rule error/auto-fix |
|-|
| image |

---

## Supplemental changes

I was also getting errors when I was running `npm run lint` due to our `@typescript-eslint` version not supporting ~~our~~ my local TS version (`5.4`).

```
G:\Coding\Spyglass [main]> npm run lint

> @spyglassmc/root@1.0.0 lint
> eslint --cache --max-warnings=0 packages/*/{src,test}/**/*.{cts,mts,ts}

=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <5.1.0

YOUR TYPESCRIPT VERSION: 5.4.5

Please only submit bug reports when using the officially supported version.

=============
```

So I updated `@typescript-eslint` to the latest major version `v7.9` (since `v7.0` was the minimum version with TS `5.4` support anyway).

---

Upgrading `@typescript-eslint` to `v7.9` led to some `Unsupported engine` warnings whenever I'd run an `npm` command:

||
|-|
|image|

I had node `v18.17.1` installed locally, but we didn't actually specify anywhere a minimum node version for the Spyglass repo as a whole.

So I went with `@typescript-eslint`'s recommended versions (listed in that warning above): `^18.18.0 || >=20.0.0`

- `^18.18.0` is LTS
- `>=20.0.0` is any modern node version

These engines felt like a reasonable requirement. Upgrading my local node to v18 LTS ([`v18.20.2`](https://nodejs.org/en/download/package-manager)) resolved these engine discrepancy warnings with `@typescript-eslint`.

---

I think after updating `@typescript-eslint` to support my local Typescript version, the `@typescript-eslint/prefer-readonly` began to emit warnings it had not previously. So I included those auto-fixes in this PR as well: [`ef9d9c1` (#1181)](https://github.com/SpyglassMC/Spyglass/pull/1181/commits/ef9d9c100ba554315c825e31663a417ada3d2433).

---

Unrelated, we forgot to run `npm i` after updating `mcdoc-cli` from `0.1.3` to `0.1.5`. So I did that to keep our `package-lock.json` up-to-date in [`d3f9713` (#1181)](https://github.com/SpyglassMC/Spyglass/pull/1181/commits/d3f9713c260d79c6791f9df8eec049646eb5fee2).


NeunEinser:
Would it be possible to add a rule for unnecessary imports as well?

TheAfroOfDoom:
> Would it be possible to add a rule for unnecessary imports as well?

I'm not sure what you mean by this specifically, but the plugin I added `eslint-plugin-import` has a bunch of other import lint rules. feel free to look through that and see if it has what you're looking for

NeunEinser:
I mean this situation.
image

Ideally, you wouldn't be allowed to commit this. It can happen quite easily when changing code and forgetting to delete these imports is quite easy.

SPGoding:
https://eslint.org/docs/latest/rules/no-unused-vars probably works for unused vars in import

.eslintrc.cjs
```diff
@@ -12,7 +12,8 @@ module.exports = {
 		"project": "./packages/**/tsconfig.json"
 	},
 	"plugins": [
-		"@typescript-eslint"
+		"@typescript-eslint",
+    "import"
```
SPGoding, line 16: inconsistent indent (two tabs on line 15, four spaces on line 16). same for line 42

package.json
```diff
@@ -3,6 +3,9 @@
 	"private": true,
 	"version": "1.0.0",
 	"license": "MIT",
+  "engines": {
+    "node": "^18.18.0 || >=20.0.0"
+  },
```
SPGoding, line 8: indent

SPGoding (commented):
The `@spyglassmc/mcdoc-cli` `version` really should just be `0.1.0-PLACEHOLDER` in `package.json` like every other package in this repository since the version is set by the release script and should not be changed manually.

.eslintrc.cjs
TheAfroOfDoom: ugh, i missed this because i turned off whitespace diffs on github

TheAfroOfDoom: nevermind that's not at all why i missed it

SPGoding (approved):

MulverineX (approved):

PR #1182, opened 2024-05-20
🔧 Remove attributed type in favor of attributes array on all types

NeunEinser:
## Motivation
The attributed type contained a single attribute with another child, which itself could be an attributed type for multiple attributes on a single type.

This can be quite annoying to deal with, especially when different processes still need to be able to access attributes, but also you want to check whether a type node is of a certain type.

With the old concept, you'd need to essentially recurse into the type tree for as long as the current node is still an attributed type.

With this, all types can simply have attributes and they can be dealt with when appropriate.

This is also more consistent with `StructTypeField` and `EnumTypeField`.

### DIFF
--- a/__snapshots__/packages/mcdoc/test-out/__fixture__/attributed_types.spec.js
+++ b/__snapshots__/packages/mcdoc/test-out/__fixture__/attributed_types.spec.js
@@ -30,13 +30,12 @@ exports['mcdoc __fixture__ attributed types 1'] = {
       "::test::NoValue": {
         "data": {
           "typeDef": {
-            "kind": "attributed",
-            "attribute": {
-              "name": "deprecated"
-            },
-            "child": {
-              "kind": "boolean"
-            }
+            "kind": "boolean",
+            "attributes": [
+              {
+                "name": "deprecated"
+              }
+            ]
           }
         },
         "subcategory": "type_alias",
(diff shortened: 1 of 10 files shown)


SPGoding (approved):

SPGoding:
Side note: would it make sense to do the same for `IndexedType` as well or are you fine with that

NeunEinser:
> Side note: would it make sense to do the same for `IndexedType` as well or are you fine with that

I think that's included, no?
```ts
export interface IndexedType extends McdocBaseType {
	kind: 'indexed'
	parallelIndices: Index[]
	child: McdocType
}

export interface McdocBaseType {
	attributes?: Attribute[]
}
```

Or did I miss a binder?

SPGoding:
I meant if you'd want to remove `IndexedType` and add the indices to the type base instead, since `IndexedType` is also recursive. Not sure if that'd make anything easier/harder for you cuz I haven't checked how you implemented things

NeunEinser:
Hmm, maybe I misunderstood indexed?

I understood it as "here is a child type, if this child resolves to a struct, use the index I give to you to index into it and return that sub type instead"

`parallelIndices` is a list already though, and so is each entry of `parallelIndices`.

So I don't really know where the recursive definition is , but maybe I am doing something wrong.

SPGoding:
> I understood it as "here is a child type, if this child resolves to a struct, use the index I give to you to index into it and return that sub type instead"

That's correct.

> So I don't really know where the recursive definition is

It's only recursive if there are multiple parallel indices (i.e. brackets) in a row. e.g. `Foo[Bar,Baz][Qux][[id]]` will be a few nested `IndexedType`s.

NeunEinser:
Yeah, I think that should be fine, as my simplify function resolves indexed types (and it _should_ be able to handle them recursively).

The issue with attributes was more that we will need to carry around attributes for later logic, so I can't simplify them away. For indexed types, I can just work with the resolved struct. I guess maybe I still need copy attributes to some resolved types.

misode (approved):
I think it makes sense to do this for attributes and not for indexed types

PR #1189, opened 2024-05-21
🔧 Add long timeout arg to unit test launch configuration

TheAfroOfDoom:
- when you're debugging these, it's not unlikely you'd be in a test longer than 30s (the default timeout)
- new default is 999999s (11.5 days)

---

- split from https://github.com/SpyglassMC/Spyglass/pull/1180


SPGoding (approved):
what happens if we stare at a breakpoint continuously for 11.5 days?

TheAfroOfDoom:
> what happens if we stare at a breakpoint continuously for 11.5 days?

GKfjTrfW8AE5HgN

PR #1190, opened 2024-05-21
🔥 Remove CommandTreeRegistry

SPGoding:
It's unnecessary abstraction that obscures things. If we ever end up having multiple versions of the tree available at the same time, whatever package's responsible for that should manage all the trees internally.


__snapshots__/packages/language-server/test-out/util/toLS.spec.js
```diff
@@ -0,0 +1,86 @@
...
+      },
+      "end": {
+        "line": 2,
+        "character": 2
+      }
+    }
+  },
+  "insertTextFormat": 2,
+  "insertTextMode": 2
+}
```
TheAfroOfDoom, line 86: looks like you need to `npm run clean` (or `clean:full`) maybe? guessing these snapshots shouldnt be showing up here

TheAfroOfDoom (approved):
curious what the initial motivation was behind storing a registry/map of trees?

also:

https://youtu.be/52LJaWDdG9c?t=26

__snapshots__/packages/language-server/test-out/util/toLS.spec.js
TheAfroOfDoom: i dont know why i felt the need to highlight the entire snapshot, sorry

SPGoding: i even double checked to make sure i didn't include them :sob:

misode (approved):

PR #1152, opened 2024-05-15
🦺 Validate command tree patch

SPGoding:
Resolves #1113.

## Prerequisites

- #1191
- #1192

## Changes

- [x] Patch `minecraft:nbt_compound_tag` parsers
- [x] Patch `minecraft:nbt_path` parsers
- [x] Patch `minecraft:nbt_tag` parsers
- [x] Patch `minecraft:resource_location` parsers
- [x] Patch `minecraft:uuid` parsers
- [x] Add necessary `permission` properties for new commands

PR #1191, opened 2024-05-21
⬆️ Upgrade dprint

SPGoding:
Split from #1152.


TheAfroOfDoom (approved):

misode (approved):
Is this just a general upgrade to improve DX, or was there a particular issue with the old version?

SPGoding:
dprint used to crash on the patch file and I think the upgrade fixes that.

PR #1192, opened 2024-05-21
🦺 Add tree patch validator

SPGoding:
Split from #1152.

Warnings will be logged during development for missing command tree patches as suggested in https://github.com/SpyglassMC/Spyglass/pull/1152#issuecomment-2115270597

image

## Prerequisites

* #1190
* #1191


MulverineX (approved):

packages/java-edition/src/mcfunction/tree/patch-validator.ts
```diff
@@ -0,0 +1,37 @@
+import type * as core from '@spyglassmc/core'
```
misode, line 1: The dash in this file name seems a bit odd, no? I would expect `patchValidator.ts`

MulverineX: Or patch_validator

misode (approved):

MulverineX:
I will actually test this later, then I'll merge it

MulverineX:
Yep, this is working! 👍

PR #1193, opened 2024-05-21
🐛 Set imported binding as an alias of target symbol

### LINKED ISSUE #1154: References to use statements aren't being resolved properly by mcdoc yet
image
image
image

The use statement itself is properly resolved to the definition, but when referencing it somewhere, you get a reference with the same name but the path of the current module.

This eventually leads to `options.context.symbols.query(ctx.doc, 'mcdoc', '::java::data::predicate::Predicate')` being resolved as a union of `(::java::data::predicate::LootCondition | [::java::data::predicate::LootCondition])` which itself is not a valid mcdoc reference.

```
use super::loot::LootCondition

type Predicate = (LootCondition | [LootCondition])
dispatch minecraft:resource[predicate] to Predicate
```

SPGoding:
Fixes #1154.

https://github.com/SpyglassMC/Spyglass/assets/15277496/221a79c2-a43b-4c87-8e4e-bed2a97012af

The file-level binding introduced by `use` is now treated as an alias of the global symbol.


packages/core/src/symbol/SymbolUtil.ts
```diff
@@ -972,7 +972,7 @@ type QueryMemberCallback = (this: void, query: SymbolQuery) => unknown
 /* istanbul ignore next */
 export class SymbolQuery {
 	readonly category: string
-	readonly path: readonly string[]
+	path: readonly string[]
 	readonly #doc: TextDocument
```
TheAfroOfDoom, line 976: what does a hash `#` prefix indicate for class properties?

packages/mcdoc/src/binder/index.ts
```diff
@@ -526,16 +539,22 @@ async function bindPath(
...
+				try {
+					query.resolveAlias().enter({
+						usage: {
+							type: 'reference',
+							node: identNode,
+							fullRange: node,
+							skipRenaming: LiteralNode.is(identNode),
+						},
+					})
+				} catch (_ignored) {
```
TheAfroOfDoom, line 552: (nitpick)

```suggestion
				} catch {
```

if we're already explaining why we do nothing with the error i don't see the need to grab the parameter

packages/mcdoc/src/binder/index.ts
```diff
@@ -247,10 +247,15 @@ function hoist(node: ModuleNode, ctx: McdocBinderContext): void {
...
 		// hoistUseStatement associates the AST node with the binding definition in the file symbol table,
-		// which may get overridden by bindUseStatement in the later stage as an reference to the imported symbol in the global symbol table.
+		// which will get overridden by bindUseStatement in the later stage as an reference to the imported symbol in the global symbol table.
 		// This way when the user tries to go to definition on the path in the use statement,
 		// they will go to the definition in the imported file.
 
+		/**
+		 * String representation of the absolute path of the imported symbol.
+		 */
+		const importedPathString = pathArrayToString(resolvePath(path, ctx))
```
TheAfroOfDoom, line 257: (nitpick)

this fits comfortably on one line if you're open to that stylistically, and still renders as JSDoc

```suggestion
		/** String representation of the absolute path of the imported symbol. */
		const importedPathString = pathArrayToString(resolvePath(path, ctx))
```

NeunEinser:
This does not fully fix #1154.

The symbol deployed to mcdoc still contains the non-aliased reference. If I delete the `visibility: SymbolVisibility.File,` in line 271, and edit my code to also do `symbol.resolveAlias()` when performing the lookup, I _do_ get the correct node with the correct typedef.

I am not sure it is desirable to declare a public alias here though, I think I would prefer it if the reference type used the resolved path, and the alias wasn't public.

NeunEinser (changes requested):
This does not fully fix #1154.

The symbol deployed to mcdoc still contains the non-aliased reference. If I delete the `visibility: SymbolVisibility.File,` in line 271, and edit my code to also do `symbol.resolveAlias()` when performing the lookup, I _do_ get the correct node with the correct typedef.

I am not sure it is desirable to declare a public alias here though, I think I would prefer it if the reference type used the resolved path, and the alias wasn't public.

packages/core/src/symbol/SymbolUtil.ts
SPGoding: [private property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties)

packages/mcdoc/src/binder/index.ts
SPGoding: didn't know you could do that

packages/mcdoc/src/binder/index.ts
MulverineX: it makes more sense to consistently have JSDoc formatted this way though, and having a line break in-between is nice for readability, since the comment color is so dark

NeunEinser: Yeah, I think I also prefer to have it always multiline.

TheAfroOfDoom: works for me!

SPGoding:
> The symbol deployed to mcdoc still contains the non-aliased reference.

I think that's solved now. I also added [tests](https://github.com/SpyglassMC/Spyglass/pull/1193/files#diff-74b13033e73582aa8ac1f6d0c1c2eeb89728f4a8cd2925e8eb9b20fbcee9938eR95) that shows the global target path is stored in the symbol table instead of the file-level alias path.

NeunEinser (approved):
Nice, this works like a charm and no cluttering the global symbol table with import aliases =)

NeunEinser:
Btw, you can use my [NeunEinser:neun](https://github.com/NeunEinser/Spyglass/tree/neun) branch which is a combination of SPG's branch and my mcdoc branch for testing and compare errors with just the [NeunEinser:mcdoc-validator](https://github.com/NeunEinser/Spyglass/tree/mcdoc-validator) branch on its own to see it working.

NeunEinser (approved):
I don't think I quite understand why alias didn't work. Is this a bug in how aliases are implemented or expected behavior when trying to declare an alias of sth not loaded yet?


If it is expected behavior, how would you use aliases in a way that avoids this problem?

All known edge cases are solved, [NeunEinser:neun](https://github.com/NeunEinser/Spyglass/tree/neun) is updated.

SPGoding:
> expected behavior when trying to declare an alias of sth not loaded yet

it's mostly this. i could still use an alias but don't actually check the existence of the target when resolving it, in which case using  a custom data is just simpler than using an alias.

packages/core/src/symbol/SymbolUtil.ts
```diff
@@ -1337,6 +1337,8 @@ export class SymbolQuery {
 			}
 			this.#symbol = result
 			this.#map = result.parentMap
+			this.#parentSymbol = result.parentSymbol
+			this.path = result.path
```
misode, line 1341: Since aliases are no longer used in this specific case, are these changes still needed? Or was this an issue that always existed before that you discovered while trying aliases?

misode (approved):
The issues on Neun's branch seem to be resolved!

packages/core/src/symbol/SymbolUtil.ts
SPGoding: It's an issue that always existed

PR #1196, opened 2024-05-22
✨ Macro Highlighting and Error Checking

### LINKED ISSUE #1122: Macro syntax highlighting
Support basic highlighting and error checking

Calverin:
Support for macro lines, highlights macro keys `$(foo)` differently from other content in the line.
Errors thrown for invalid characters in a macro key, a blank macro, missing parenthesis, and a macro line (line starting with a $) that doesn't have a macro in it.

~Notes also written describing default colorizer colors.~

Closes #1122


MulverineX (approved):

MulverineX (approved):

packages/mcfunction/src/node/macro.ts
```diff
@@ -0,0 +1,66 @@
...
+			type: 'mcfunction:macro',
+			range: core.Range.get(range),
+			children: [MacroChildNode.mock(range)],
+		}
+	}
+}
+
+export interface MacroOptions {
+	type?: 'macro' | 'other' | 'sign'
+	colorTokenType?: core.ColorTokenType
```
misode, line 28: Storing the color token in the AST feels wrong. The AST is supposed to represent the structure of the document. I haven't reviewed the rest of the PR yet, but imo a command like `$say $(message)` should look a bit like this (simplified AST):
```yaml
type: mcfunction:macro
children:
- type: mcfunction:macro_start
  value: "$"
- type: mcfunction:macro_raw
  value: 'say '
- type: mcfunction:macro_key
  children:
  - type: string
    value: message
```

MulverineX: It seems like there's precedent for this in lots of other nodes?

Calverin: The colorTokenType code was taken from `LiteralNode` and it's done the same way there. As for storing the type it's done more for organizational reasons and it's done the same way for `ResourceLocation`s

misode: The colorizer would look at the type of the nodes and assign colors based on that. Color tokens in nodes are currently only used for some literal types. Not for distinquishing between the color of parts like these in a command: `$(`, `say `, `message`. Those shouldn't all have the same `macro_child` node type.

MulverineX: Actually we came up with a way to improve upon this, once its pushed hopefully that's an adjustment that can resolve this

Calverin: I've removed the `colorTokenType` parameter and just the nodes are now just colored based on their type (which is just literal or string)

misode: Looks like an improvement, but the `options.type` just seems like a secondary type. Why do they share the same `macro_child` node type instead of being 3 different AST node types?

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,153 @@
...
+export function macro(): core.Parser<MacroNode> {
+	return (
+		src: core.Source,
+		ctx: core.ParserContext,
+	): core.Result<MacroNode> => {
+		const start = src.cursor
+		const children: MacroChildNode[] = []
+
+		// Handle the starting '$'
+		if (src.peek() == '$' && src.peek(2) !== '$(') {
```
misode, line 17: Why do you need to check `src.peek() == '$'`? This parse function is only called when this condition is already true. And what is the purpose of checking for `$(` in the beginning of a macro line? It would be an invalid command but the parser shouldn't care about that.

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,153 @@
...
+	}
+}
+
+function parse(
+	src: core.Source,
+	ctx: core.ParserContext,
+	txt: String,
+	beginning: number,
+	wasMacro: Boolean,
+): MacroChildNode {
```
misode, line 101: What does this function do? It's not very descriptive, including the `txt`, `beginning`, `wasMacro` (which isn't even used) parameters.

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,153 @@
...
+			src.skip()
+			children.push(ans)
+		}
+
+		// Handle the rest of the line
+
+		let beginning = src.cursor
+		let txt = src.readUntil('$', '\r', '\n')
+		do { // Do-while because we always want it to work the first time
+			let wasMacro = false
```
misode, line 33: I'll be honest, the `txt` and `wasMacro` variables storing random state make the logic very confusing to follow.

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,153 @@
...
+			const child: MacroChildNode = parse(src, ctx, txt, beginning, wasMacro)
+			if (child.options.type === 'macro') {
+				wasMacro = true
+			}
+			// Handle a non-macro $ in the middle of the line
+			let subMacro: MacroChildNode | undefined
```
misode, line 39: Why do we need to handle non-`$` in the middle of the line? Those could just be part of the command. What are sub macros?

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,153 @@
...
+			if (
+				child.options.type === 'other' && child.value?.indexOf('$') !== -1
+			) {
+				const parts: String[] = child.value?.split('$') ?? []
```
misode, line 43: Why is this splitting the string result of the previous `parse` call? What would the parts even be in a normal command?

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,153 @@
...
+				children.push(subMacro)
+			}
+			// Prepare for the next block
+			beginning = src.cursor
+			if (wasMacro) {
+				txt = src.readUntil('\r', '\n', '$')
+			} else {
+				txt = src.readUntil('\r', '\n', ')')
+			}
+		} while (txt.length > 0)
```
misode, line 73: Instead of checking the length of a string, use something like `src.canReadInLine()`

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,153 @@
...
+
+		// No actual macro
+		if (children.length < 3) {
```
misode, line 76: This is a very hacky way to check whether there were macro keys. Consider filtering the children list and checking their type. This would presumably also fail for the line `$$(cmd)`, which is valid.

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,153 @@
...
+		}
+
+		// Return the result
+		const ans: MacroNode = {
+			type: 'mcfunction:macro',
+			range: core.Range.create(start),
+			children: children,
+		}
+
+		ans.range.end = src.cursor
```
misode, line 90: You can just return the node with something like `core.Range.create(start, src)`

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,153 @@
...
+				core.Range.create(beginning, src.cursor),
+			)
+		}
+		if (check.length === 0) { // No key
+			ctx.err.report(
+				localize('expected', localize('macro-key')),
+				core.Range.create(beginning, src.cursor),
+			)
+		}
+		if (txt.charAt(txt.length - 1) !== ')') { // Missing parenthesis
```
misode, line 123: Avoid using raw strings when parsing, this can cause problems when used with index maps.

misode (commented):
The parsing looks really messy and overcomplicated. I've tried leaving comments at several places where I'm not sure why a choice was made, but in general I think the parsing logic could do with a large rewrite.

packages/mcfunction/src/parser/macro.ts
Calverin: In a command like `$scoreboard players set $mode settings $(value)` the `$mode` would mess with the existing parsing, a sub macro is a macro that gets caught in the rest of the command as a by-product of that problem and is extracted with the following code

misode: What if you have the line `$tellraw $(players) ["$ Hello everyone! $"]`?

packages/mcfunction/src/parser/macro.ts
Calverin: In a normal command it wouldn't be split because it wouldn't contain a `$`, it's designed only to handle that edge case

packages/mcfunction/src/parser/macro.ts
Calverin: Not sure what this is in reference to?

packages/mcfunction/src/parser/macro.ts
Calverin: But the line is already read in the `src.readUntil()`s so it could've already reached the end of the line, hence why it's checking that the string isn't empty

packages/mcfunction/src/parser/macro.ts
Calverin: image
Currently works okay

packages/mcfunction/src/parser/macro.ts
```diff
@@ -76,8 +74,15 @@ export function macro(): core.Parser<MacroNode> {
...
 		} while (nextChunk.length > 0)
 
-		// No actual macro
-		if (children.length < 3) {
+		// A line with no macros is invalid
+        let hasMacro = false
+        for (const child of children) {
+            if (child.type === 'mcfunction:macro_child/macro') {
+                hasMacro = true
+                break
```
misode: This can probably be compacted to `children.some(c => c.type === 'mcfunction:macro_child/macro')`

packages/mcfunction/src/parser/macro.ts
misode: You renamed the variables but imo the issue is still there.

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,161 @@
...
+			children.push(child)
+			if (subMacro) { // Sub-macro added here because colorizer
+				children.push(subMacro)
+			}
+			// Prepare for the next block
+			beginning = src.cursor
+			if (wasMacro) {
+				nextChunk = src.readUntil('\r', '\n', '$')
+			} else {
+				nextChunk = src.readUntil('\r', '\n', ')')
```
TheAfroOfDoom, line 73: i see a handful of `\r` and `\n` checks, you can probably just wrap the whole parser in `stopBefore('\r', '\n')` and remove these explicit checks

we do that for the `argument` parser:

https://github.com/SpyglassMC/Spyglass/blob/2b74e1ff633f56db439c6084d1bb72c16f536a88/packages/java-edition/src/mcfunction/parser/argument.ts#L93-L98

packages/mcfunction/src/parser/macro.ts
misode: Yeah I get that it "works" okay, but I'm more concerned in how it works. (I edited my previous comment afterwards)

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,161 @@
...
+/**
+ * Takes in a chunk of text read from the source and returns a MacroChildNode of type macro or other.
+ */
+function parseMacroChild(
+	src: core.Source,
+	ctx: core.ParserContext,
+	chunk: String,
+	beginning: number,
+): MacroChildNode {
+	if (chunk.substring(0, 2) === '$(') { // This is a macro key
```
misode, line 111: This function essentially does two completely unrelated things: validating a macro key (first if branch) and returning a node for the surrounding part of a macro line.

Calverin: Correct

packages/mcfunction/src/colorizer/index.ts
```diff
@@ -18,8 +16,20 @@ export function register(meta: core.MetaRegistry) {
...
+		macroChild,
+	)
+	meta.registerColorizer<MacroChildNode>(
+		'mcfunction:macro_child/other',
+		macroChild,
+	)
+	meta.registerColorizer<MacroChildNode>(
+		'mcfunction:macro_child/sign',
+		macroChild,
+	)
```
TheAfroOfDoom, line 30: would be nice if we had a `registerColorizers` where we could just pass in an array of node types

packages/mcfunction/src/parser/macro.ts
```diff
@@ -14,15 +14,13 @@ export function macro(): core.Parser<MacroNode> {
...
-				type: 'mcfunction:macro_child/sign',
-				range: core.Range.create(start, start + 1),
-				value: '$',
-			}
-			src.skip()
-			children.push(sign)
+		const sign: MacroChildNode = {
+			type: 'mcfunction:macro_child/sign',
+			range: core.Range.create(start, start + 1),
+			value: '$',
```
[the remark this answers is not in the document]
Calverin: As far as I can tell there's no built-in way to specifically colorize the first character in a line. Even using the `fillGap()` method I'm using for the macro keys it only works up until the first macro, at which point it reverts to the rest of the line being highlighted the same color as the $ sign

packages/mcfunction/src/parser/macro.ts
TheAfroOfDoom: you should generally be using `src.peek(<number>)` instead. if you don't use `IndexMap`s then the biggest thing is this PR won't work with the backslash continuation logic

packages/mcfunction/src/parser/macro.ts
TheAfroOfDoom: something you can do to make a complicated function more readable is to split the two underlying things its doing into 2 separate functions. (or more if you're feeling daring 😚)

TheAfroOfDoom (commented):
leaving a couple minor comments, ill take another more in-depth after Misode's reviews are mostly done!

packages/mcfunction/src/colorizer/index.ts
Calverin: Don't think that's in my department 😆

packages/mcfunction/src/parser/macro.ts
Calverin: image
This is what the subtree looks like

packages/mcfunction/src/parser/macro.ts
Calverin: Tried to explain things some more and reorganize things a bit

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,108 @@
...
+			range: core.Range.create(src.cursor),
+			children: [],
+		}
+
+		// Skip the starting '$'
+		ans.children.push({
+			type: 'mcfunction:macro_child/macro',
+			range: core.Range.create(src.cursor, src.cursor + 1),
+			value: '$',
+		})
```
misode, line 24: Reusing the same AST node as macro keys is somehow even more of a hack. I suggest either adding the old node type back or colorizing the `$` without there being a node for it.

packages/mcfunction/src/node/macro.ts
```diff
@@ -0,0 +1,61 @@
...
+			range: core.Range.get(range),
+			children: [MacroChildNode.mock(range)],
+		}
+	}
+}
+
+export interface MacroChildNode extends core.AstNode {
+	type:
+		| 'mcfunction:macro_child/macro'
+		| 'mcfunction:macro_child/other'
```
misode, line 29: These two node types shouldn't share the same interface. You should have `MacroGapNode` and `MacroVariableNode` and then `MacroChildNode` can be a union of those two types.

packages/mcfunction/src/node/macro.ts
```diff
@@ -0,0 +1,61 @@
...
+	value?: String
```
misode, line 30: In TypeScript, always use `string` instead of `String`

MulverineX: I'm sure he knows that 😝, just a typo

Calverin: Unfortunate habit from Java

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,122 @@
...
+		let hasMacroKeys = false
+
+		while (src.canReadInLine()) {
+			src.skipUntilOrEnd(core.LF, core.CR, '$')
+			if (src.peek(2) === '$(') {
+				// Add the gap before this macro key
+				const gap = src.sliceToCursor(start)
+				if (gap.length > 0) {
+					children.push({
+						type: 'mcfunction:macro/gap',
```
NeunEinser, line 42: So `macro/gap` is everything between macro arguments?

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,122 @@
...
+				if (src.peek() != ')') {
+					// Macro key was not closed
+					ctx.err.report(
+						localize('expected', localeQuote(')')),
+						core.Range.create(keyStart, src.cursor),
+					)
+				} else if (src.cursor <= keyStart) {
+					// Encountered $()
+					ctx.err.report(
+						localize('expected', localize('macro-key')),
```
NeunEinser, line 62: No change to en.json, so presumably missing translation for `macro-key`

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,122 @@
...
+						core.Range.create(start, src.cursor + 1),
+					)
+				}
+				const key = src.sliceToCursor(keyStart)
+				const matchedInvalid = key.replace(/[a-zA-Z0-9_]*/, '')
+				if (matchedInvalid.length > 0) {
+					ctx.err.report(
+						localize(
+							'parser.resource-location.illegal',
```
NeunEinser, line 71: Illegal resource location?

I think this should either be a different key or the existing key should be renamed to reflect its intended use better.

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,122 @@
...
+					value: src.sliceToCursor(start),
+					path: [src.sliceToCursor(start)],
+				})
+			}
+		}
+
+		// A line with no macros is invalid
+		if (!hasMacroKeys) {
+			ctx.err.report(
+				localize('expected', localize('macro')),
```
NeunEinser, line 109: Missing translation key?

Also shouldn't it be sth like `at-least-one-macro-argument`?

packages/mcfunction/src/colorizer/macro.ts
```diff
@@ -0,0 +1,32 @@
...
+	MacroSignNode,
+	MacroVariableNode,
+} from '../node/macro'
+
+export const macroChild: core.Colorizer<
+	MacroGapNode | MacroVariableNode | MacroSignNode
+> = (node, ctx) => {
+	const color = node.type === 'mcfunction:macro/gap'
+		? 'string'
+		: 'literal'
```
NeunEinser, line 14: Not sure `literal` is a good name, and there are some duplicate color mappings already. `macro_sign` could work I guess, as that's also how you call the node. Could be sth sth string interpolation marker as well, but that's quite long.

I think the other color names you are using somewhat fit.

packages/mcfunction/src/parser/macro.ts
MulverineX: Yes

packages/mcfunction/src/parser/macro.ts
MulverineX: There's other missing English locale strings too, should probably do them all in-batch. We also need an issue for it

packages/mcfunction/src/parser/macro.ts
MulverineX: The key should be renamed, as it's useful for both of these usecases

packages/mcfunction/src/colorizer/macro.ts
MulverineX: Color names that resolve to the same color, aren't all that necessary, become a little unruly if we have too many

packages/mcfunction/src/parser/macro.ts
SPGoding: Gap makes me think of whitespaces. Would something like `macro/raw_text` be a more fitting name?

Calverin: Yes

packages/mcfunction/src/colorizer/macro.ts
Calverin: It's not a matter of naming it's just a matter of using built-in colorizer colors (didn't feel the need to add more duplicates). This isn't referenced anywhere else so I don't think it's an issue

packages/mcfunction/src/parser/macro.ts
Calverin: I had "other" originally but went with Misode's suggestion of "gap". Not personally a fan of "raw_text" but maybe another alternative?

packages/locales/src/locales/en.json
```diff
@@ -208,5 +208,8 @@
   "unquoted-string": "an unquoted string",
   "unsorted-keys": "Unsorted keys",
   "uuid": "a UUID",
-  "vector": "a vector"
+  "vector": "a vector",
+  "parser.macro.key": "a macro key",
+  "parser.macro.at-least-one": "at least one macro",
```
NeunEinser, line 213: Shouldn't it be "at least one macro argument"?

"Macro" is more the whole file, or maybe it could be argued the line is a macro, referring to a macro argument as just "a macro" seems strange to me, at least I have never heard it used that way.

packages/mcfunction/src/parser/macro.ts
MulverineX: maybe `macro/text_body`?

MulverineX (approved):

packages/mcfunction/src/parser/macro.ts
TheAfroOfDoom: > There's other missing English locale strings too, should probably do them all in-batch.

i kinda strongly disagree with this, if we're actively referencing a new locale string in a PR we should be diligent about adding it

(easier to do this with people being active with reviews)

MulverineX: alright, well looks like Calverin added them anyway

packages/locales/src/locales/en.json
```diff
@@ -208,5 +208,8 @@
   "unquoted-string": "an unquoted string",
   "unsorted-keys": "Unsorted keys",
   "uuid": "a UUID",
-  "vector": "a vector"
+  "vector": "a vector",
+  "parser.macro.key": "a macro key",
+  "parser.macro.at-least-one": "at least one macro argument",
+  "parser.macro.illegal": "Illegal key character: “%0%”"
```
SPGoding, line 214: This should be scoped under `mcfunction.` and the file should be ordered by key lexicographically.

packages/mcfunction/src/colorizer/macro.ts
```diff
@@ -0,0 +1,39 @@
...
+	const tokens: core.ColorToken[] = []
+	tokens.push(
+		core.ColorToken.create(
+			core.Range.create(node.range.start, node.range.start + 1),
+			'literal',
+		),
+	) // Dollar Sign
+	for (const child of node.children) {
+		if (child.type === 'mcfunction:macro/other') {
+			tokens.push(core.ColorToken.create(child.range, 'string')) // Other stuff is Orange
```
SPGoding, line 14: I don't know how I feel about describing semantic color tokens with names of colors (same with the added comments in `packages/core/src/processor/colorizer/Colorizer.ts`). It is not the language server's concern what color this text should be and it is beyond the language server's capabilities to set those colors. The language server should choose a token based on purely the semantic meaning of the text and select the existing token with the closest alignment (in this case, `string` is a good choice due to macro line's similarity to template strings in other programming languages).

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,101 @@
...
+export function macro(): core.Parser<MacroNode> {
+	return (
+		src: core.Source,
+		ctx: core.ParserContext,
+	): core.Result<MacroNode> => {
+		const ans: MacroNode = {
+			type: 'mcfunction:macro',
+			range: core.Range.create(src.cursor),
+			children: [],
+			path: [],
```
SPGoding, line 17: What is `path` for?

misode: I was also wondering this. I just tried removing them and the tests still pass and it still works when trying it out locally.

Calverin: It was to avoid this error we were getting whenever a character was changed
`[Error - 12:32:06 AM] [Service] [getSignatureHelp] Failed for “file:///c:/Users/mulve/AppData/Roaming/PrismLauncher/instances/1.21/.minecraft/saves/New%20World/datapacks/enchantment_reference/data/funny_mobs/functions/funny_mob/spawn.mcfunction” #38 TypeError: Cannot read properties of undefined (reading 'length')
    at getOptions3 (c:\Users\mulve\Documents\misode-spyglass\packages\vscode-extension\dist\server.js:46280:54)
    at c:\Users\mulve\Documents\misode-spyglass\packages\vscode-extension\dist\server.js:46237:21
    at Service.getSignatureHelp (c:\Users\mulve\Documents\misode-spyglass\packages\vscode-extension\dist\server.js:38357:24)
    at c:\Users\mulve\Documents\misode-spyglass\packages\vscode-extension\dist\server.js:47801:24
[Error - 12:32:36 AM] [Service] [getSignatureHelp] Failed for “file:///c:/Users/mulve/AppData/Roaming/PrismLauncher/instances/1.21/.minecraft/saves/New%20World/datapacks/enchantment_reference/data/funny_mobs/functions/funny_mob/spawn.mcfunction” #71 TypeError: Cannot read properties of undefined (reading 'length')`
I traced it down to the path not existing for the nodes

SPGoding: Can we fix the mcfunction signature provider instead of adding an unnecessary `path` property to macro nodes?

SPGoding: Specifically the type assertion here is no longer safe since macro nodes can also be selected here. (Why didn't comment nodes cause us this problem?)

https://github.com/SpyglassMC/Spyglass/blob/9d415a9b8d06e9367fa2e0434f291aeae8e57766/packages/java-edition/src/mcfunction/signatureHelpProvider.ts#L70-L72

misode: I see, the issue is that `java-edition/src/mcfunction/signatureHelpProvider.ts` `getSelectedCommandNode` is incorrectly assuming that every child it finds will be a `CommandNode` (by casting to it). The fix would be to check the node type in `signatureHelpProvider` and only continue when it's actually a command node.

SPGoding: NINJA'D

Calverin: So should I remove the path stuff and just leave the bug to be fixed in another PR?

packages/mcfunction/src/colorizer/macro.ts
Calverin: So you want me to remove the comments in Colorizer as well?

SPGoding (approved):

packages/locales/src/locales/en.json
```diff
@@ -75,6 +75,9 @@
   "mcfunction.parser.entity-selector.player-name.too-long": "Player names cannot be longer than %0% characters",
   "mcfunction.parser.eoc-unexpected": "Expected more arguments",
   "mcfunction.parser.leading-slash": "a leading slash %0%",
+  "mcfunction.parser.macro.at-least-one": "at least one macro argument",
+  "mcfunction.parser.macro.illegal": "Illegal key character: “%0%”",
+  "mcfunction.parser.macro.key": "a macro key",
```
misode, line 80: Inconsistent use of "argument" and "key"

Calverin: Well the latter two errors are related to the contents within the parenthesis, consistently referred to as the key, and the former is related to the lack of any macro variable or argument, which pertains to the discussion we had as to what that is called but it is different from the consistently defined key.

NeunEinser: Would `invalid macro argument` or `malformed macro argument` describe this error message?

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,103 @@
...
+				const matchedInvalid = key.replace(/[a-zA-Z0-9_]*/, '')
+				if (matchedInvalid.length > 0) {
+					ctx.err.report(
+						localize(
+							'mcfunction.parser.macro.illegal',
+							matchedInvalid.charAt(0),
+						),
+						core.Range.create(keyStart, src.cursor),
+					)
+				}
```
TheAfroOfDoom, line 65: can we factor this code section into a helper function? purely for readability -- something like `validateMacro()`

packages/mcfunction/src/colorizer/macro.ts
```diff
@@ -0,0 +1,39 @@
...
+			core.Range.create(node.range.start, node.range.start + 1),
+			'literal',
+		),
+	) // Dollar Sign
+	for (const child of node.children) {
+		if (child.type === 'mcfunction:macro/other') {
+			tokens.push(core.ColorToken.create(child.range, 'string'))
+		} else {
+			const start = child.range.start
+			const end = child.range.end
```
TheAfroOfDoom, line 17:
```suggestion
			const { start, end } = child.range
```

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,103 @@
...
+				src.skip()
+				ans.children.push({
+					type: 'mcfunction:macro/argument',
+					range: core.Range.create(start, src.cursor),
+					value: key,
+				})
+				start = src.cursor
+				hasMacroArgs = true
```
TheAfroOfDoom, line 73: is this flag _always_ set to `true` if the `if (src.peek(2) === '$(')` is triggered?

if so, let's move it to the top right after the condition. it being at the bottom of the block right now is making me question whether or not its actually always set to `true` when the block is triggered

TheAfroOfDoom (changes requested):

packages/mcfunction/src/parser/macro.ts
TheAfroOfDoom: its very long and looks like it exclusively checks for invalid macros, it being refactored into a helper would make it clear that everything _else_ in the `macro` function is for actual parsing

packages/locales/src/locales/en.json
NeunEinser: Yeah, if we have consensus on a preferred wording, it makes sense to change now.

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,103 @@
+import * as core from '@spyglassmc/core'
+import { localeQuote, localize } from '@spyglassmc/locales'
+import type { MacroNode } from '../node/index.js'
+
+/**
+ * Parse a macro line.
+ */
+export function macro(): core.Parser<MacroNode> {
```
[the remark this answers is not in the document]
TheAfroOfDoom: if you want some help with what kind of test cases i can be around to give you advice in voice

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,103 @@
...
+		}
+
+		// Skip the starting '$'
+		src.skip()
+
+		let start = src.cursor
+		let hasMacroArgs = false
+
+		while (src.canReadInLine()) {
+			src.skipUntilOrEnd(core.LF, core.CR, '$')
```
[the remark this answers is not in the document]
SPGoding: `macro` is concerned with the entire macro line so i think it makes sense to have it handle `LF`/`CR` inside of it.

packages/mcfunction/src/parser/macro.ts
SPGoding: `validateMacroArgument()` would match the terminology we use here better.

packages/mcfunction/test/parser/macro.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+
+describe('mcfunction parser macro()', () => {
+	const cases: { content: string }[] = [
+		{ content: '$say $(message)' },
+		{ content: '$tag @s add $(tag)_40' },
+		{ content: '$scoreboard players set $mode settings $(value)' },
+		{ content: '$tellraw $(player) $(text)' },
+		{ content: '$tellraw $(players) ["$ Hello everyone! $"]' },
+		{ content: '$summon cow $(x).1 $(y)30.0 $(z).0' },
+	]
```
TheAfroOfDoom, line 17: this is great! nice job 😄 

we should also add at least one case for each type of error we can report so far.

something like this should cover all our bases:

```ts
{ content: '$say $(foo bar' },
{ content: '$say $() bar' },
{ content: '$say $(invalid.key) bar' },
{ content: '$say no macro argument specified' },
```

TheAfroOfDoom: the motivation behind this is so that we can be confident we don't have a regression in functionality if, for whatever reason, we need to modify the `macro` parser in the future (not that I expect us to need to)

packages/mcfunction/src/colorizer/macro.ts
```diff
@@ -0,0 +1,39 @@
+import * as core from '@spyglassmc/core'
+import type { MacroNode } from '../node/macro'
+
+export const macro: core.Colorizer<MacroNode> = (node, ctx) => {
```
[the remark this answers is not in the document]
TheAfroOfDoom: (from discord)

> Got to admit I don't understand the Colorizer unit tests at all so I'm going to avoid that for now

admittedly I haven't touched the colorizer before nor its unit tests, and it looks simple enough so its probably fine to skip unit tests for it 👍

TheAfroOfDoom (changes requested):
one more thing in regards to the unit test cases, but otherwise LGTM!

packages/mcfunction/src/parser/macro.ts
```diff
@@ -0,0 +1,123 @@
...
+						range: core.Range.create(start, src.cursor),
+						value: other,
+					})
+					start = src.cursor
+				}
+				// Parse the macro key
+				const key = validateMacroArgument(src, ctx, start)
+				if (key === core.Failure) {
+					return core.Failure
+				}
```
TheAfroOfDoom, line 43: i think these early returns are a breaking change. i reverted the changes to `parser/macro.ts` found in [`331919e` (#1196)](https://github.com/SpyglassMC/Spyglass/pull/1196/commits/331919ec14bb6fff77bf852598f2dc8dc4775ad8#diff-de781f026023e7c4048802b049dce94c3eff97dba7987568181956055fe0ce28), and i was able to run the server.

tests also still pass fine. not sure what the initial motivation was behind this? i'm not getting any TS errors when i remove those changes

TheAfroOfDoom (changes requested):

TheAfroOfDoom (approved):
looks great! thanks for putting up with my unit test obsession 🤪

PR #1201, opened 2024-05-24
🧑‍💻 Enable source maps in stack traces

SPGoding:
image

### DIFF
--- a/packages/vscode-extension/.vscodeignore
+++ b/packages/vscode-extension/.vscodeignore
@@ -2,6 +2,7 @@
 */**
 
 !dist/*.js
+!dist/*.js.map
 !resource/*.json
 !CHANGELOG.md
 !icon.png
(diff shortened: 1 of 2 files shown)


NeunEinser (approved):

TheAfroOfDoom (approved):
holy shit

misode (approved):
Works!

SPGoding:
I just added another commit to enable this for production as well. According to [Node.js doc](https://nodejs.org/docs/v20.13.1/api/cli.html#--enable-source-maps)

> Note, enabling source maps can introduce latency to your application when Error.stack is accessed. If you access Error.stack frequently in your application, take into account the performance implications of --enable-source-maps.

We do not access `Error.stack` unless there is an actual crash, so I think this should be negligible.

NeunEinser (approved):

MulverineX (approved):

PR #1209, opened 2024-05-27
✨ Add support for @n

### LINKED ISSUE #1204: Support for the new @n entity selector

Calverin:
Basic support and unit tests for the new `@n` entity selector!

Closes #1204

### DIFF
--- a/packages/java-edition/src/mcfunction/node/argument.ts
+++ b/packages/java-edition/src/mcfunction/node/argument.ts
@@ -120,21 +120,8 @@ export namespace EntitySelectorArgumentsNode {
 		)
 	}
 }
-export const EntitySelectorVariables = ['a', 'e', 'p', 'r', 's', 'n']
+export const EntitySelectorVariables = ['a', 'e', 'p', 'r', 's', 'n'] as const
 export type EntitySelectorVariable = typeof EntitySelectorVariables[number]
-export function removeUnavailableSelectorVariables(ctx: core.ContextBase) {
-	const release = ctx.project['loadedVersion'] as
-		| ReleaseVersion
-		| undefined
-	// Remove @n if the version is lower than 1.21
-	if (
-		EntitySelectorVariables.includes('n') && release &&
-		ReleaseVersion.cmp(release, '1.21') < 0
-	) {
-		EntitySelectorVariables.pop()
-		EntitySelectorAtVariables.pop()
-	}
-}
 export namespace EntitySelectorVariable {
 	/* istanbul ignore next */
 	export function is(value: string): value is EntitySelectorVariable {
@@ -150,6 +137,13 @@ export namespace EntitySelectorAtVariable {
 	export function is(value: string): value is EntitySelectorAtVariable {
 		return EntitySelectorAtVariables.includes(
 			value as EntitySelectorAtVariable,
+		)
+	}
+
+	export function filterAvailable(ctx: core.ContextBase) {
+		const release = ctx.project['loadedVersion'] as ReleaseVersion | undefined
+		return EntitySelectorAtVariables.filter(variable =>
+			!(variable === '@n' && release && ReleaseVersion.cmp(release, '1.21'))
 		)
 	}
 }
--- a/packages/java-edition/src/mcfunction/node/argument.ts
+++ b/packages/java-edition/src/mcfunction/node/argument.ts
@@ -143,7 +143,8 @@ export namespace EntitySelectorAtVariable {
 	export function filterAvailable(ctx: core.ContextBase) {
 		const release = ctx.project['loadedVersion'] as ReleaseVersion | undefined
 		return EntitySelectorAtVariables.filter(variable =>
-			!(variable === '@n' && release && ReleaseVersion.cmp(release, '1.21'))
+			!(variable === '@n' && release &&
+				ReleaseVersion.cmp(release, '1.21') < 0)
 		)
 	}
 }
--- a/packages/java-edition/src/mcfunction/node/argument.ts
+++ b/packages/java-edition/src/mcfunction/node/argument.ts
@@ -120,7 +120,7 @@ export namespace EntitySelectorArgumentsNode {
 		)
 	}
 }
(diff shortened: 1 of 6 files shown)


misode (changes requested):
The parser needs to check the current version and error if necessary

packages/java-edition/src/mcfunction/parser/argument.ts
```diff
@@ -724,7 +724,8 @@ function selector(): core.Parser<EntitySelectorNode> {
 						: undefined
 					predicates = variable === '@e' ? ['Entity::isAlive'] : undefined
 					single = variable
-						? variable === '@p' || variable === '@r' || variable === '@s'
+						? variable === '@p' || variable === '@r' ||
+							variable === '@s' || variable === '@n'
```
TheAfroOfDoom, line 728:
```suggestion
						? ['@p', '@r', '@s', '@n'].includes(variable)
```

packages/java-edition/src/mcfunction/parser/argument.ts
```diff
@@ -445,10 +444,10 @@ function entity(
 				children: [res],
 			}
 
-			if (core.StringNode.is(res)) {
-				ans.playerName = res
-			} else if (EntitySelectorNode.is(res)) {
+			if (EntitySelectorNode.is(res)) {
 				ans.selector = res
+			} else if (core.StringNode.is(res)) {
+				ans.playerName = res
```
misode, line 450: I can't immediately see why this would break, but what's the reason for switching these around?

packages/java-edition/src/mcfunction/node/argument.ts
```diff
@@ -119,8 +120,21 @@ export namespace EntitySelectorArgumentsNode {
...
+	const release = ctx.project['loadedVersion'] as
+		| ReleaseVersion
+		| undefined
+	// Remove @n if the version is lower than 1.21
+	if (
+		EntitySelectorVariables.includes('n') && release &&
+		ReleaseVersion.cmp(release, '1.21') < 0
+	) {
+		EntitySelectorVariables.pop()
+		EntitySelectorAtVariables.pop()
```
misode, line 135: This is really ugly and probably a bad idea to modify a global constant. Instead consider adding a `filterEntitySelectorVariables(ctx): string[]` pure function (or similarly named). This would also break in the future if we end up supporting multiple versions or allow switching the loaded version without a full vscode reload.

packages/java-edition/src/mcfunction/parser/argument.ts
```diff
@@ -1321,14 +1330,27 @@ function selector(): core.Parser<EntitySelectorNode> {
...
+		(res, _, ctx) => {
+			const selector = res.children.find(core.LiteralNode.is)!.value.slice(1)
+			const release = ctx.project['loadedVersion'] as
+				| ReleaseVersion
+				| undefined
+			// Invalid @ selector (including @n before 1.21)
+			if (
+				!(EntitySelectorVariable.is(selector)) ||
+				(release && ReleaseVersion.cmp(release, '1.21') < 0 &&
+					selector === 'n')
```
misode, line 1342: This should re-use the same function as the completer to determine the currently allowed selector variables, although this whole logic should be moved to the `selectorPrefix` function (as explained in my later comment)

packages/java-edition/src/mcfunction/parser/argument.ts
```diff
@@ -690,6 +689,19 @@ function range(
 	)
 }
 
+function selectorPrefix(): core.InfallibleParser<core.LiteralNode> {
+	return (src: core.Source): core.LiteralNode => {
```
misode, line 693: This parse function should check the allowed selector variables, for reference look at `mcfunction/src/parser/literal.ts`

misode (changes requested):

packages/java-edition/src/mcfunction/parser/argument.ts
Calverin: It was just to avoid any @ selectors getting picked up as potential player names (or UUIDs as they had been in the past)

packages/java-edition/src/mcfunction/node/argument.ts
Calverin: The problem is the completer uses the list for suggestions, and I couldn't figure out another way to modify the suggestions without editing the list. Any suggestions @misode?

misode (approved):

packages/java-edition/src/mcfunction/node/argument.ts
```diff
@@ -138,6 +139,13 @@ export namespace EntitySelectorAtVariable {
 			value as EntitySelectorAtVariable,
 		)
 	}
+
+	export function filterAvailable(ctx: core.ContextBase) {
+		const release = ctx.project['loadedVersion'] as ReleaseVersion | undefined
+		return EntitySelectorAtVariables.filter(variable =>
+			!(variable === '@n' && release && ReleaseVersion.cmp(release, '1.21'))
```
SPGoding, line 146: am i tripping?
```suggestion
			!(variable === '@n' && release && ReleaseVersion.cmp(release, '1.21') < 0)
```

misode: Not sure how this happened, thanks for catching this!

misode (approved):

MulverineX (approved):

packages/java-edition/src/mcfunction/node/argument.ts
```diff
@@ -119,7 +120,7 @@ export namespace EntitySelectorArgumentsNode {
 		)
 	}
 }
-export const EntitySelectorVariables = ['a', 'e', 'p', 'r', 's'] as const
+export const EntitySelectorVariables = ['a', 'e', 'p', 'r', 's', 'n'] as const
```
SPGoding, line 123: (pretty minor) I think we could remove the `export ` from `EntitySelectorVariables` and `EntitySelectorAtVariables` and add a comment to note that `The filterAvailable() function should be used to get a list of available selector variables` to avoid any future mistakes?

SPGoding (approved):

PR #1221, opened 2024-05-30
🐛 Re-enable undeclared symbols linter

### LINKED ISSUE #1220: Valid resource locations to unknown symbols don't show an error
In the command `give @s invalid:item`, no error is shown, despite the `invalid:item` not existing in the symbol table. This affects all resource locations.

misode:
Fixes #1220 

Also improves the resource location completer so it no longer suggests the current or other undeclared symbols.
|Before|After|
|---|---|
|image|image|


MulverineX (approved):

SPGoding (approved):
i have no idea why it was turned off

PR #1231, opened 2024-06-01
✨ Fully support mcdoc defined resources, refactor java-edition binder

### LINKED ISSUE #1169: Legacy folders have been renamed to match the registry name
In 1.21 the legacy folders like `tags/items` have been renamed to `tags/item`. Since 24w21a this covers all legacy directory names, not just tags.

### LINKED ISSUE #1218: Use the `minecraft:resource` mcdoc dispatcher for JSON file validation
Right now the mcdoc identifiers are hardcoded for each file category, it should use the mcdoc dispatcher to figure out the right identifier.

MulverineX:
Fixes #1169, Resolves #1218, ~~includes some mcdoc-cli work that can be largely ignored~~


packages/java-edition/src/binder/index.ts
```diff
@@ -166,17 +169,50 @@ export function* getRels(
...
 
 	const rels = getRels(uri, ctx.roots)
+
+	const { customResources } = ctx.config.env
+
+	if (
+		customResources.length > 0 &&
+		ctx.project['customResourcesDefined'] === undefined
+	) {
+		ctx.project['customResourcesDefined'] = 'done'
```
misode, line 184: What's the reason for this state? This feels highly error prone.

packages/java-edition/src/json/checker/index.ts
```diff
@@ -1,66 +1,25 @@
...
 }
 
 export function register(meta: core.MetaRegistry) {
-	meta.registerChecker<json.JsonFileNode>('json:file', file)
+	meta.registerChecker<json.JsonNode>('json:array', entry)
+	meta.registerChecker<json.JsonNode>('json:boolean', entry)
+	meta.registerChecker<json.JsonNode>('json:null', entry)
+	meta.registerChecker<json.JsonNode>('json:number', entry)
+	meta.registerChecker<json.JsonNode>('json:object', entry)
+	meta.registerChecker<json.JsonNode>('json:string', entry)
```
misode, line 24: Looks like a merge conflict resolved the wrong way. It should only handle `json:file` nodes now.

packages/json/src/checker/index.ts
```diff
@@ -2,6 +2,109 @@ import type * as core from '@spyglassmc/core'
 import * as mcdoc from '@spyglassmc/mcdoc'
 import { type JsonNode, JsonPairNode } from '../node/index.js'
 
+const Options: (
+	ctx: core.CheckerContext,
+) => mcdoc.runtime.checker.McdocCheckerOptions<JsonNode> = (ctx) => ({
```
misode, line 7: This is a function, so it should be called `function getOptions(ctx: core.CheckerContext)`

packages/core/src/service/Config.ts
```diff
@@ -41,6 +64,15 @@ export interface EnvConfig {
...
 	dependencies: string[]
+	/**
+	 * **Experimental feature, breaking changes could occur.**
+	 *
+	 * Custom resources, currently only works for `minecraft:resource` JSON definitions dispatched from mcdoc.
+	 */
+	customResources: [
+		filePath: string,
+		config: CustomResourceConfig,
+	][]
```
misode, line 75: I've said it before, but I really don't like us exposing these kinds of configs at this stage. We're trying to get a minimal working version out. Things like this should really be discussed in an issue first and put on the backlog.

Especially because it is a java-edition specific config that we're putting in the core config basically.

misode (changes requested):
I won't make an issue about it now, but in the future please try to avoid mixing unrelated mcdoc-cli changes in a PR. Apart from that, I'm not a fan of the `customResources` config.

packages/java-edition/src/json/checker/index.ts
MulverineX: I couldn't resolve this without errors, I'm not sure how to handle this properly, I'd appreciate a fix, can't hop in VC this week.

packages/json/src/checker/index.ts
MulverineX: Gotcha, I'll take care of that later today

packages/core/src/service/Config.ts
MulverineX: As I said in #contributors, this field will be unsupported. 

As for the java-edition specific thing, I'll add the `pack` field here.

packages/java-edition/src/binder/index.ts
MulverineX: Its an optimization measure, if you can figure out a better way to handle this please let me know :)

packages/core/src/service/Config.ts
misode: > if it doesn't work it isn't blocking

We don't want to ship broken config code. If something isn't working, in my eyes it's blocking.

> As for the java-edition specific thing, I'll add the pack field here later today.

Not sure what you mean with this and how it addresses my concerns.

MulverineX: I disagree. Here, all of this code is pretty flat and straight forward, I'll do some testing later today to make sure it doesn't blow up.

`pack: 'datapack'` would be the only value that would actually do anything for now, with `pack: 'java_resource_pack` being implemented eventually.

MulverineX:
In the future I'll avoid including mcdoc-cli stuff in unrelated PRs.

packages/core/src/service/Config.ts
```diff
@@ -26,6 +26,29 @@ export interface Config {
...
+	 */
+	extname: '.json'
+	/**
+	 * The pack type of the dispatched resource. Only `datapack` is supported for now.
+	 */
+	pack: 'datapack'
+	/**
+	 * The version range of the dispatched resource.
+	 */
+	since?: `1.${number}`
```
NeunEinser, line 45: If we get 10 patch versions, how do we differentiate 1.x.1 and 1.x.10 with this format?

MulverineX: Well this is leftover from before my PR, I think the idea here was to not track minor versions in this file at all? I'd be up for changing this though

packages/core/src/service/Config.ts
MulverineX: oh right, this is already how its setup, it uses `pack`. no changes here are needed

misode (changes requested):
(not sure why my previous comment went through as approved)

packages/java-edition/src/binder/index.ts
misode: Okay I see. I don't think this is necessary, especially since tha large majority of people won't even use this config

packages/java-edition/src/binder/index.ts
```diff
@@ -10,130 +10,132 @@ import type {
...
+// 	version?: ScopedVersion,
+// ) {
+// 	return JSONResource('java_resource_pack', dispatcher, pathOrVersion, version)
+// }
+
+type CategoryDef = {
+	category: FileCategory | TagFileCategory
+	extname: string
+	pack: 'datapack' | 'java_resource_pack'
+} & ScopedVersion
```
TheAfroOfDoom: i think this is _slightly_ easier to read

```suggestion
interface CategoryDef extends ScopedVersion = {
	category: FileCategory | TagFileCategory
	extname: string
	pack: 'datapack' | 'java_resource_pack'
}
```

packages/java-edition/src/binder/index.ts
```diff
@@ -166,17 +169,38 @@ export function* getRels(
 	return undefined
 }
 
-export function dissectUri(uri: string, ctx: ContextBase) {
+export function dissectUri(uri: string, ctx: UriBinderContext) {
 	const regex =
 		/^data\/([^\/]+)\/((?:tags\/|worldgen\/)?[a-z_]+)\/(.*)(\.(?:mcfunction|json))$/
 
 	const rels = getRels(uri, ctx.roots)
+	const { customResources } = ctx.config.env
```
TheAfroOfDoom, line 171: shorthand property accessing 🥳

packages/java-edition/src/binder/index.ts
```diff
@@ -166,17 +169,38 @@ export function* getRels(
...
+				for (let i = 0; i < subFolders.length - 1; i++) {
+					const customCategory = `${match[2]}/${
+						subFolders.slice(0, i + 1).join('/')
+					}`
+
+					def = Categories.get(customCategory)
+
+					// we don't break here because we want to find the most specific category
+				}
+			}
```
TheAfroOfDoom, line 196: we don't need the `else` block if the `if` is doing a `continue`. this lets us avoid unnecessary indentation

```suggestion
			if (subFolders.length === 0) {
				continue
			}
			
			for (let i = 0; i < subFolders.length - 1; i++) {
				const customCategory = `${match[2]}/${
					subFolders.slice(0, i + 1).join('/')
				}`

				def = Categories.get(customCategory)

				// we don't break here because we want to find the most specific category
			}
```

TheAfroOfDoom: actually, we dont need the explicit `subFolders.length === 0` check at all because the for-loop will be a no-op if it's an empty list

packages/java-edition/test/json/binder/index.spec.ts
```diff
@@ -1,11 +1,11 @@
-import { ContextBase } from '@spyglassmc/core'
+import { ContextBase, UriBinderContext } from '@spyglassmc/core'
```
TheAfroOfDoom, line 1: `ContextBase` appears unused. we can remove the import

packages/mcdoc-cli/src/generate/index.ts
```diff
@@ -20,870 +20,222 @@ export async function generate(
...
-
-				if (child.type === 'resource_location') {
-					/* @ts-ignore */
-					child.namespace = _child.namespace
-					/* @ts-ignore */
-					child.path = _child.path
-				}
-
-				if (Object.hasOwn(_child, 'isOptional')) {
+	const project_path_uri = pathToFileURL(project_path).toString() + '/'
```
TheAfroOfDoom, line 31: why is everything in this file snake_case 😭

TheAfroOfDoom (changes requested):
from @misode 

> I won't make an issue about it now, but in the future please try to avoid mixing unrelated mcdoc-cli changes in a PR. Apart from that, I'm not a fan of the `customResources` config.

tbh i'd strongly prefer if we split this up into multiple PRs. its hard to tell what changes are for what issues :L

packages/java-edition/src/binder/index.ts
```diff
@@ -10,130 +10,132 @@ import type {
 import { fileUtil, TaggableResourceLocationCategories } from '@spyglassmc/core'
 import { ReleaseVersion } from '../dependency/index.js'
 
+type ScopedVersion = { since?: ReleaseVersion; until?: ReleaseVersion }
+
+function JSONResource(
```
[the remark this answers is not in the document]
misode: Fixed

packages/java-edition/src/binder/index.ts
```diff
@@ -10,130 +10,132 @@ import type {
...
+	pack: 'datapack' | 'java_resource_pack',
+	dispatcher: FileCategory | TagFileCategory,
+	pathOrVersion: string | ScopedVersion,
```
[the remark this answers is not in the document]
misode: I think I simplified it a bit in the last commit

packages/java-edition/src/binder/index.ts
```diff
@@ -10,130 +9,124 @@ import type {
...
-			{
-				category: 'worldgen/template_pool',
-				extname: '.json',
-				since: '1.16',
-			},
-		],
+	const ans = new Map<string, CategoryDef>([
+		// Pre-1.21
+		dataPackResource('advancement', 'advancements', {
+			since: '1.16',
```
misode, line 59: This map is still missing a lot of resources (like wolf variants, damage types, trim materials, etc)

packages/java-edition/src/binder/index.ts
```diff
@@ -10,130 +10,132 @@ import type {
...
+) {
+	let _version: ScopedVersion = {}
+
+	if (typeof pathOrVersion !== 'string') {
+		_version = pathOrVersion
+	} else if (version) {
+		_version = version
+	}
+	return [
+		typeof pathOrVersion === 'string' ? pathOrVersion : dispatcher,
```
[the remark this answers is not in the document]
misode: Cleaned this up

package-lock.json
```diff
@@ -5634,9 +5634,9 @@
 			}
 		},
 		"node_modules/flatted": {
-			"version": "3.2.5",
-			"resolved": "https://registry.npmjs.org/flatted/-/flatted-3.2.5.tgz",
-			"integrity": "sha512-WIWGi2L3DyTUvUrwRKgGi9TwxQMUEqPOPQBVi71R96jZXJdFskXEmf54BoZaS1kknGODoIGASGEzBUYdyMCBJg==",
+			"version": "3.3.1",
+			"resolved": "https://registry.npmjs.org/flatted/-/flatted-3.3.1.tgz",
+			"integrity": "sha512-X8cqMLLie7KsNUDSdzeN8FYK9rEt4Dt67OsG/DNGnYTSDBG4uFAJFBnUeiV+zCVAvwFy56IjM9sH51jVaEhNxw==",
```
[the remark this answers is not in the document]
MulverineX: I tried using `flatted` for mcdoc-cli, so it touched that. Nothing broke so I think I'm just going to leave this

packages/core/src/service/Config.ts
```diff
@@ -26,6 +26,29 @@ export interface Config {
...
+	/**
+	 * The version range of the dispatched resource.
+	 */
+	until?: `1.${number}`
```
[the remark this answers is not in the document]
MulverineX: I fixed this :+1:

packages/mcdoc-cli/src/generate/index.ts
MulverineX: `mcdoc-cli` is my project, I'm going to keep it like this for now, we can discuss this another time like in v4.1

package-lock.json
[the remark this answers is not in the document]
TheAfroOfDoom: uhh, `package-lock.json` should always be in sync with `package.json`

if you are not touching dependencies then the lock file should be 0 diff

packages/core/src/service/Config.ts
MulverineX: Looks like there's usages of minor versions via `ReleaseVersion` already, so I'll resolve this. Added some resources that include the minor version too.

misode (approved):

packages/java-edition/src/binder/index.ts
```diff
@@ -178,18 +178,14 @@ export function dissectUri(uri: string, ctx: UriBinderContext) {
...
-					def = Categories.get(customCategory)
-
-					// we don't break here because we want to find the most specific category
-				}
+			for (let i = 0; i < subFolders.length - 1; i++) { // if subFolders is empty, this loop will not run
+				const customCategory = `${match[2]}/${
+					subFolders.slice(0, i + 1).join('/')
+				}`
+
+				def = Categories.get(customCategory)
```
TheAfroOfDoom, line 207: wait. isn't this loop ultimately just the same as doing this?

```ts
const customCategory = `${match[2]}/${subFolders.join('/')}`
def = Categories.get(customCategory)
```

we're never doing anything with the `def` assignment mid-loop so it will functionally only ever do stuff with the final iteration of the loop

packages/java-edition/src/binder/index.ts
```diff
@@ -10,130 +9,145 @@ import type {
...
+	category: FileCategory | TagFileCategory,
+	path: string | ScopedVersion,
+	version?: ScopedVersion,
+): [string, CategoryDef] {
+	if (typeof path !== 'string') {
+		// Shift parameters if path is a scoped version
+		version = path
+		path = category
+	}
+	return [path, { pack, category, extname: '.json', ...version }]
```
TheAfroOfDoom, line 33: instead of relying on positional-argument jank, can we just make this an object so the parameter position doesn't matter?

```ts
function resource(
	pack: 'data_pack',
	{ category, path, version }: {
		category: FileCategory | TagFileCategory,
		path: string,		
		version?: ScopedVersion,
	}
): [string, CategoryDef] {
	return [path, { pack, category, extname: '.json', ...version }]
```

TheAfroOfDoom (changes requested):
these are kinda minor, but they're the last stylistic/readability comments i have!

packages/java-edition/src/binder/index.ts
misode: I think the reason for the positional arguments (I wouldn't say it's jank), it to keep the function calls short in the map below.

TheAfroOfDoom: ahh that's fair

PR #1232, opened 2024-06-02
🐛 Fix semantic token builder not coloring each line of a color token

TheAfroOfDoom:
# Summary

I'll have some upcoming PRs relating to fixing the colorizer across multiple lines (due to backslash continuation)

I wanted to add tests for this, but it wouldn't be great with snapshots since the return type of `semanticTokens()` is just an array of integers (kinda meaningless to the human eye)

https://github.com/SpyglassMC/Spyglass/blob/9bb9dc1bb0c70f178a6b5efac20816346411d621/packages/language-server/src/util/toLS.ts#L321-L325

---

@SPGoding do you remember what file type you were using to colorize tokens across multiple lines prior to backslash continuation?
  - relevant commit is from 2 years ago: https://github.com/SpyglassMC/Spyglass/commit/e15cf81ec9060f2b7d9055a0faa4c67fba736fd1

---

## Preview

|| before | after |
|-|-|-|
| commands w/line continuation | image | image |
| comments + macro lines | image|image |

---

some noticeable colorization errors still remaining are:

1. commas being miscolored when near line continuation + strings
this is due to `syntax-mcfunction` colorizing strings line-wise, and the comma is caught between the quotes on that line

2. start/end brackets being miscolored/not matching up across line continuation
this is due to `syntax-mcfunction`, as disabling the extension makes the brackets yellow/match properly. see below image:

||
|-|
|image|

removing `syntax-mcfunction` as an extension dependency is not an option, so we'll instead probably have to explicitly make tokens for them in Spyglass to override `syntax-mcfunction`'s incorrectly coloring

### DIFF
--- a/__snapshots__/packages/language-server/test-out/util/toLS.spec.js
+++ b/__snapshots__/packages/language-server/test-out/util/toLS.spec.js
@@ -1,0 +1,29 @@
+exports['semanticTokens Tokenize "foo" 1'] = [
+  {
+    "deltaLine": 0,
+    "deltaStartChar": 0,
+    "length": 100,
+    "tokenType": 0,
+    "tokenModifiers": 0
+  }
+]
+
+exports['semanticTokens Tokenize "foo↓bar" 1'] = [
+  {
+    "deltaLine": 0,
+    "deltaStartChar": 0,
+    "length": 100,
+    "tokenType": 0,
+    "tokenModifiers": 0
+  }
+]
+
+exports['semanticTokens Tokenize "foo↓bar↓qux" 1'] = [
+  {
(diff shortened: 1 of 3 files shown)


SPGoding (approved):
oFf bY OnE

SPGoding:
> commas being miscolored when near line continuation + strings

is there an example of this?

TheAfroOfDoom:
> > commas being miscolored when near line continuation + strings
> 
> is there an example of this?

my wording is kinda wrong (in an attempt to be concise), but its basically when you line-continuate mid-string

image

misode (approved):

PR #1234, opened 2024-06-04
✅ Add mcdoc runtime tests

misode:
Depends on #1187

### DIFF
--- a/packages/mcdoc/test/runtime/checker.spec.ts
+++ b/packages/mcdoc/test/runtime/checker.spec.ts
@@ -30,15 +30,237 @@ describe('mcdoc runtime checker', () => {
 			],
 		},
 		{
-			name: 'struct { test: string }',
+			name: 'struct { foo: string, bar: [double] }',
 			type: {
 				kind: 'struct',
 				fields: [
-					{ kind: 'pair', key: 'test', type: { kind: 'string' } },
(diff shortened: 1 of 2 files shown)


packages/mcdoc/test/runtime/checker.spec.ts
```diff
@@ -0,0 +1,455 @@
...
+			return { kind: 'any' }
+		}
+	}
+
+	for (const { name, type, values, init } of suites) {
+		describe(`typeDefinition ${localeQuote(name)}`, () => {
+			for (const value of values) {
+				it(`with value ${JSON.stringify(value)}`, () => {
+					const errors: McdocCheckerError<JsValue>[] = []
+					const project = mockProjectData()
```
NeunEinser, line 392: This looks useful!

NeunEinser (approved):
This looks nice for a start. I'd probably want to write some more in-depth tests for some things interacting in weird ways, but this is good for a start.

I think it still makes sense if I write some tests too, as I have thought _a lot_ about all kinds of weird edge cases in the past week, and I am probably the most into it rn :)

But yeah we should merge this as soon as possible. Thanks for your help, really appreciate it :) <3

packages/mcdoc/test/runtime/checker.spec.ts
misode: Yep, it's a pretty useful utility

PR #1236, opened 2024-06-04
🐛 Fix parent and key special index accessors

misode:


packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -965,18 +965,13 @@ export function simplify<T>(
...
+						if (typeof entry != 'string' && entry.keyword === 'parent') {
 							possibilities = possibilities.map(n => {
 								const node = n.parents.pop()
 								return { parents: n.parents, node }
 							})
 						} else if (
-							entry === '%key' ||
-							(typeof entry != 'string' && entry.keyword === 'key')
+							typeof entry != 'string' && entry.keyword === 'key'
 						) {
```
NeunEinser, line 975: Can this be one line now or does the formatter not like that?

NeunEinser (approved):

packages/mcdoc/src/runtime/checker/index.ts
NeunEinser: Oh also it probably should be `typeof entry !== 'string'` (instead of `!=` with one `=`)

misode: Formatter doesn't like it. I know it also bugged me a bit. Issue would go away when we split this in `simplifyIndexed`.

NeunEinser: Not sure what `resolveIndex` would do, and what the diference would be, regardless splitting it off would be a good idea.

To avoid bigger conflicts with my wip experiment to simplify the graph structure, it would be good to avoid bigger structural changes for now though.

PR #1237, opened 2024-06-05
✨ Check mcdoc numeric value ranges

misode:


packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -571,7 +571,36 @@ function condenseErrorsAndFilterSiblings<T>(
...
+		} else {
+			const rangesErrors = validDefinitions.map(d => {
+				return d.errors
+					.filter((e): e is RangeError<T> => e.kind === kind)
+					.reduce((a, b) => ({
+						kind,
+						node: a.node,
+						ranges: [NumericRange.intersect(a.ranges[0], b.ranges[0])],
+					}))
+			})
```
NeunEinser, line 594: Can you add this test:
```rs
type A = (
    struct {
        foo: int @0..5
    }
    | struct {
        foo: int @0..15
    }
)
```

My hunch is that this breaks there. I am not sure I fully understand this logic, but it seems to rely on that you are only condensing the immeidate child where the error happened.

If the above works try this instead:
```rs
type A = (
    struct {
        foo: (int @0..5 | int @20..25) 
    }
    | struct {
        foo: int @0..15
    }
)
```

NeunEinser: The issue is, I don't think the validDefinitions.map is really the correct approach.

You could for example also have this:
```rs
type A = (
    struct {
        foo: (int @0..5) 
    }
    | struct {
        foo: int @4..6
    }
)
```
Now, you kind of expect a merged numeric range which says "between 0 and 6" if both fail.

Or even this, though one could argue that this is a definition issue in this case and should be defined differently:
```rs
(int @0..5 | 5..7)
```

misode: > Now, you kind of expect a merged numeric range which says "between 0 and 6" if both fail.

That may be a better error in the future, but I wouldn't strictly consider it being a bug. But right now it gives an error like `Expected an int at least 0 and at most 5 or at least 0 and at most 6`. When you add some brackets this is still correct.

misode: Either way, both of the cases you mentioned "work", even though the error message is a bit ambiguous without brackets

NeunEinser: Ohh, I see I actually misread the code a bit, but there is a different issue here.

You are actually just merging _all_ range errors from _all_ valid definitions into one, regardless of which node they are about.

What I actually expected for the once I mentioned above is not that you report one error like you do, but a duplicated error with two conflicting ranges.

Here is an actual issue though:
```rs
struct {
    foo: int @ 0..5,
    bar: int @ 4..6
}
```
When I test this against a value were both of them produce an error, you combine both errors into one, which is incorrect, for example:

```js
exports['mcdoc runtime checker typeDefinition “struct { foo: int @ 0..5, bar: int @ 4..6 }” with value {"foo":6,"bar":2} 1'] = [
  {
    "kind": "number_out_of_range",
    "node": {
      "originalNode": 6,
      "inferredType": {
        "kind": "literal",
        "value": {
          "kind": "double",
          "value": 6
        }
      }
    },
    "ranges": [
      {
        "kind": 0,
        "min": 4,
        "max": 5
      }
    ]
  }
]
```
This now says `Expected an int between 4 and 5` for the value of `foo`, and it doesn't report any error for `bar`, because you combine all range errors into one.

You need to keep track of which node the error was reported for, I usually do this by doing a reference check on `originalNode`. With my tree rework, this will actually be cleaner as we guranteed only have each runtime value once in the tree, so these reference checks become semantically correct.

If you don't want to handle the combination of errors from different defintions for the same key here, I think it's good to leave a //TODO.

packages/mcdoc/src/type/index.ts
```diff
@@ -27,14 +26,46 @@ export namespace NumericRange {
...
+			range.max !== undefined && (RangeKind.isRightExclusive(range.kind)
+				? val >= range.max
+				: val > range.max)
 		) {
 			return false
 		}
 		return true
 	}
+
+	export function intersect(a: NumericRange, b: NumericRange): NumericRange {
```
NeunEinser, line 38: I don't know if you actually need the `intersect` function btw, I initially interpreted it more like a union (confirmation bias here, as I thought that's the only function that should be needed).

In which case would you expect to range errors for one child to be thrown that need to be intersected? The only reason for two range errors to be about the same child I can think of is if they are different members of a union and maybe need to be combined because they have overlap.

So yeah what's the case the logic would break without this?

packages/mcdoc/src/runtime/checker/index.ts
misode: Thanks for the explanation! I'll pick this back up after merging the tree rework.

NeunEinser: Technically the most correct way would probably be sth like this:

- Collect all range errors where a range error happens for a specific node in all validDefinitions
  - Combine these range errors from into one
    - Ideally, merging different ranges into one if they have overlap rather than just making them all into a list
  - Report the resulting list of errors
- Filter all the node that have a non-common range error (A range error for a runtime node that only happens in some of the validDefinitions, but not all of them)
  - If nothing is left, report a `sometimes_range_error` kind of error for other nodes
  - Otherwise, remove all definitions that have an uncommon range error

I'll leave it up to you how much of this you want to implement in this PR though, I think it's enough for an initial merge if the bug I mentioned above is fixed.

packages/mcdoc/src/type/index.ts
misode: Yeah while I was inplementing this I couldn't think of any way the intersect is necessary, because we don't have an `&` mcdoc type. But now that we have it I don't see a reason to get rid of it 🤷

Issue #932, opened 2022-03-31
Autocomplete JSON values when halfway


kittech0:
When there is already a string in json file, it doesnt autocompete 
What should do:
![image](https://user-images.githubusercontent.com/73485099/161093438-4d690d31-2434-40a5-abd7-f65daf64fec4.png)
But it doesnt
![image](https://user-images.githubusercontent.com/73485099/161093482-55d15763-ebfd-4449-aa41-17c5b303403a.png)
becuz to make it auto complete again i need to remove string and starts typing then

MulverineX:
@kittech0 does pressing backspace then typing in `p` again make the autocomplete show up again? If so this isn't a bug

PR #1238, opened 2024-06-05
✨ Attach mcdoc type info and JSON completer

### LINKED ISSUE #1226: Add mcdoc based completers for `runtime` doc use (JSON/NBT, etc)
See also #947

This issue is more about providing a general interface for mcdoc-based completers that can be reused for JSON and NBT. This may be done at once to have a test case or separately

misode:
A rough implementation to attach type info. Adds JSON completer and simple hover information.

Resolves #1226 and #932

image
image
image
image


packages/mcdoc/src/runtime/completer/index.ts
```diff
@@ -0,0 +1,61 @@
+import type * as core from '@spyglassmc/core'
+import { TypeDefSymbolData } from '../../binder/index.js'
+import type { McdocType, StructTypePairField } from '../../type/index.js'
+
+export function getFields(
+	typeDef: core.DeepReadonly<McdocType>,
```
NeunEinser, line 6: attachtypeInfo calls this with a simplified mcdoc type, so you shouldn't need to handle struct spread operators here

packages/mcdoc/src/runtime/completer/index.ts
```diff
@@ -0,0 +1,61 @@
...
+					}
+					return []
+				})
+		default:
+			return []
+	}
+}
+
+export function getValues(
+	typeDef: core.DeepReadonly<McdocType>,
```
NeunEinser, line 30: Same here, no need tho handle references, for instance.

packages/json/src/node/index.ts
```diff
@@ -1,7 +1,22 @@
...
+}
+export namespace JsonFileNode {
+	/* istanbul ignore next */
+	export function is(obj: object | undefined): obj is JsonFileNode {
+		return (obj as JsonFileNode | undefined)?.type === 'json:file'
+	}
+}
+
+interface JsonBaseNode {
+	typeDef?: mcdoc.McdocType
```
NeunEinser, line 17: This should be a SimplifiedMcdocType

packages/mcdoc/src/runtime/completer/index.ts
misode: `getValues` is also called for an empty array, in which case the `typeDef.item` field is used, which isn't simplified yet (because the checker doesn't give any more information that that)

packages/json/src/node/index.ts
misode: I sort of prefer working with `McdocType` everywere and think the simplified types should be left mostly internal to the checker. We can still just choose not to handle certain kinds in for example the completer.

packages/mcdoc/src/runtime/completer/index.ts
misode: I don't think I do? Just for type safety I return `[]` in the flatmap when encountering it.

packages/json/src/node/index.ts
NeunEinser: I definitely prefer types that show exactly what kind of data you can expect, no more no less.

I intentionally provide a simplified mcdoctype in the interface of attachTypeInfo, for exactly this reason, to tell the user that this is the only kind of types that it can expect.

packages/mcdoc/src/runtime/completer/index.ts
NeunEinser: Oh, I read `if (field.kind === 'spread')` and then sth with a curly brace which returns sth.

Bad formatting imo, but if the formatter/styling rules for this repo don't enforce curly braces around any if, I guess it's fine. Really hard to read like this though, especially because the return is on the same line as the if.

packages/mcdoc/src/runtime/completer/index.ts
NeunEinser: Hmm, I see good point.

Would it be possible to manually call `simplify` for the completer when encountering this case? If you have a list of dispatchers for example, you would ideally be able to provide a completion for it.

I also kind of dislike the reference resolution being duplicated here, I think ideally you would make sure that this is always a simplified mcdoctype.

misode: That is definitely an option. I agree the duplicated logic isn't ideal.

misode: I've put a TODO for this, because I don't want this part of the code to use the unstable simplify function yet

NeunEinser: I think by now it should be pretty stable, but not very important either.

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -504,7 +504,26 @@ export function typeDefinition<T>(
...
+				options.attachTypeInfo(node.node.originalNode, def.typeDef)
+				attached.add(node.node.originalNode)
+			}
+		}
+		for (const child of node.children) {
+			for (const node of child.possibleValues) {
+				attachTypeInfo(node)
+			}
+		}
+	}
```
NeunEinser, line 522: The set and TODO shouldn't really be neccessary anymore.

It should be fine to attach a union type containing all `node.validDefinitions`.

misode: Cool! I'll try working on this

misode: Ah shit 😄 

image

PR #1245, opened 2024-06-08
✨ Implement mcdoc attributes and JSON tag files

### LINKED ISSUE #1162: Update symbol categories
The list of categories used by the symbol table is outdated. All the new file categories (new data driven resources) and some of the new registries that are used in those files.

https://github.com/SpyglassMC/Spyglass/blob/ad778e38026c1cadc82aac43329a19b69eb71686/packages/core/src/symbol/Symbol.ts#L21
https://github.com/SpyglassMC/Spyglass/blob/ad778e38026c1cadc82aac43329a19b69eb71686/packages/core/src/symbol/Symbol.ts#L119

### LINKED ISSUE #1217: Add mcdoc support for JSON tag files
For this, the `tag_entry` mcdoc attribute will have to be implemented first.

### LINKED ISSUE #1225: Add attribute handling to mcdoc
Proposed interface by Misode:

```ts
export interface AttributeHandler<C> {
  // parses the attribute value after the =
  config: (value: AttributeValue) => C
  // runs on every node during the checking process
  checkNode<T>: (node: RuntimeNode<T>, config: C, ctx: AttributeHandlerContext) => ValidationError<T>
  // Can be used to filer a union, or add stuff to a type, make all fields in a truct optional, etc
  mapType: (types: McdocType, config: C, ctx: AttributeHandlerContext) => McdocType
}
```

misode:
- [x] Closes #1225
- [x] Closes #1217
- [x] Closes #1162

### Implemented attributes
- [x] `since`
- [x] `until`
- [x] `id` (missing support for `tags=required`)
- [x] `command_argument` (fails for arguments that require properties)
- [x] `nbt`
- [x] `nbt_path`
- [x] `command` (has some completion and hover issues)

### Remaining issues
- `id` without category doesn't work yet
- `id` doesn't infer the `minecraft:` default namespace in all cases (see loot table `type` field)
- `nbt` and `nbt_path` don't run the mcdoc checker yet


MulverineX (approved):

packages/java-edition/src/json/checker/index.ts
```diff
@@ -1,18 +1,50 @@
...
+
 export const file: core.Checker<json.JsonFileNode> = (node, ctx) => {
 	const child = node.children[0]
 	if (ctx.doc.uri.endsWith('/pack.mcmeta')) {
-		return json.checker.definition('::java::pack::Pack')(child, ctx)
+		const type: mcdoc.McdocType = {
+			kind: 'reference',
+			path: '::java::pack::Pack',
+		}
+		return json.checker.index(type)(child, ctx)
```
NeunEinser, line 34: This seems a bit more messy, any reason to remove the `definition` shorthand?

packages/java-edition/src/json/checker/index.ts
```diff
@@ -1,18 +1,50 @@
...
+			const type = createTagDefinition(parts.category.slice(4))
+			return json.checker.index(type)(child, ctx)
+		}
+		const type: mcdoc.McdocType = {
+			kind: 'dispatcher',
+			registry: 'minecraft:resource',
+			parallelIndices: [{ kind: 'static', value: parts.category }],
 		}
-		return json.checker.resource(parts.category)(child, ctx)
+		return json.checker.index(type)(child, ctx)
```
NeunEinser, line 47: Same here for `resource` shorthand

packages/java-edition/src/mcfunction/index.ts
```diff
@@ -70,6 +71,84 @@ export const initialize = (
...
+						)
+						return
+					}
+					if (src.canRead()) {
+						ctx.err.report(
+							localize('mcdoc.runtime.checker.trailing'),
+							core.Range.create(src.cursor, src.skipRemaining()),
+						)
+					}
+					node.children = [child]
```
NeunEinser, line 122: Can this cause issues / overwrite sth? Would it be better to do this?
```ts
node.children ??= []
node.children.push(child)
```

packages/java-edition/src/mcfunction/index.ts
```diff
@@ -70,6 +71,84 @@ export const initialize = (
...
+					// TODO: fix completer and checker inside commands
+					const src = new core.Source(node.value, node.valueMap)
+					const command = mcf.command(tree, parser.argument)(src, ctx)
+					if (src.canRead()) {
+						ctx.err.report(
+							localize('mcdoc.runtime.checker.trailing'),
+							core.Range.create(src.cursor, src.skipRemaining()),
+						)
+					}
+					node.children = [command]
```
NeunEinser, line 146: same here about overwriting existing children

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -832,21 +828,20 @@ function checkShallowly<T>(
...
+				config,
+				simplifiedInferred,
+				options.context,
+			) === false
+		) {
+			assignable = false
 		}
+	})
+	if (!assignable) {
+		errors.push({ kind: 'internal', node: runtimeNode })
```
NeunEinser, line 844: I'm not sure why you are introducing `internal` here. In general, it would be better to report a more specific error, maybe have the attribute not return a boolean here, but a McdocRuntimeCheckerError instead.

packages/java-edition/src/index.ts
```diff
@@ -116,6 +118,59 @@ export const initialize: core.ProjectInitializer = async (ctx) => {
...
+					ctx.logger.warn(`Invalid mcdoc attribute for "until": ${config}`)
+					return true
+				}
+				return ReleaseVersion.cmp(release, config as ReleaseVersion) < 0
+			},
+		},
+	)
+	mcdoc.runtime.registerAttribute(
+		meta,
+		'deprecated',
```
NeunEinser, line 151: When is this used instead of `since` / `until`?

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -514,9 +493,20 @@ export function typeDefinition<T>(
 		for (const def of node.validDefinitions) {
 			if (!attached.has(node.node.originalNode)) {
 				options.attachTypeInfo(node.node.originalNode, def.typeDef)
+				handleAttributes(
+					def.typeDef.attributes,
+					options.context,
+					(handler, c) => {
+						const attacher = handler.attachString?.(c, options.context)
+						if (attacher) {
+							options.stringAttacher(node.node.originalNode, attacher)
```
NeunEinser, line 502: I'm not sure this is the correct approach as this is problematic for multiple valid definitions. I suppose we could just say calling this function multiple times is okay and let the handler deal with it in some way?

Not sure how to handle this though, it doesn't really make sense to have multiple string parsers at the same time, maybe it would be best to just not do anything at all when there are multiple conflicting valid definitions that have an attachString method defined?

packages/java-edition/src/mcfunction/index.ts
```diff
@@ -70,6 +71,84 @@ export const initialize = (
...
+							localize('mcfunction.parser.eoc-unexpected'),
+							node,
+						)
+						return
+					}
+					if (src.canRead()) {
+						ctx.err.report(
+							localize('mcdoc.runtime.checker.trailing'),
+							core.Range.create(src.cursor, src.skipRemaining()),
+						)
```
NeunEinser, line 120: I think it's bad design to report an error here.

Ideally, there should be a `filter` function defined which reports an appropriate error.

Was the intention of `internal` errors to facilitate sth like this with custom error reporting, and custom error condensing after my current PR is done? I could see that, as runtime can't really know the kinds of errors that can be reported.

I wouldn't call it `internal` I think, maybe more like `custom`, and ofc it shouldn't be created in runtime checker.

packages/mcdoc/src/runtime/checker/index.ts
misode: I don't want the mcdoc checker to show an error here, because when the string attacher runs it will report its own (better) `undeclaredSymbol` warning. But I do still need the error condenser to discard possible types that have these errors indicated by attributes (mostly to support the `prefix=!` use case)

packages/java-edition/src/mcfunction/index.ts
misode: I refactored this later where the logic is now in only one place in the checker, and it only runs once per node during the `attachTypeInfo`. We don't need to support the use case where multiple string parsers run on the same node, that's never going to work.

packages/java-edition/src/index.ts
misode: `show_item` hover event has some of these

packages/mcdoc/src/runtime/checker/index.ts
misode: The attached set makes sure it only runs once. In #1293 I added a comment, but I don't think it makes much sense to run multiple string parsers on the same node. It means you're probably missing a required type used by a dispatcher.

packages/java-edition/src/json/checker/index.ts
misode: No particular reason 🤷 I don't think it's that messy to supply a mcdoc type.

packages/java-edition/src/json/checker/index.ts
misode: Dispatching `minecraft:resource` is probably only going to happen here, so I don't see why the logic should be moved to the `json` package. If anything it's the `java-edition` package that should define how the `minecraft:resource` dispatcher works.

packages/java-edition/src/mcfunction/index.ts
misode: I don't think it's bad design. The `internal` basically means "I as the attribute will handle errors in this type, but if you ever need to decide between multiple types, you can throw me away"

For complicated parsers it's much nicer to let them report their own errors, rather than them all having to go through the mcdoc runtime checker.

packages/java-edition/src/index.ts
NeunEinser: Ah, so it's essentially when it's still valid in vanilla mc but deprecated?

packages/java-edition/src/mcfunction/index.ts
NeunEinser: Well, internal errors right now are only reported by the id attribute right now, so there actually is no filtering at all for any of the other complex parsers.

> For complicated parsers it's much nicer to let them report their own errors, rather than them all having to go through the mcdoc runtime checker.

I think, ideally all the parsing would happen when intitially encountering the node already, including reporting custom errors (as the parser would do normally, as nice as possible, not just a generic thing for the entire node). For error condensing, custom errors could be handled like you are doing it with internal right now, it would just be a bit tricky if we have comflicting custom errors from multiple definitions, but at this point we could just report a generic `invalid value` or something.

If you report errors here, and do the parsing as I suggested before, this implies to parse two times. So, there should be some way in error reporting to report custom errors. Maybe custom errors could just have a report function attached or something like that.

packages/java-edition/src/json/checker/index.ts
NeunEinser: Well, I didn't really see the point of changing it in the first place I guess. I also don't think it's a huge issue, but I have seen a few inline mcdoc type creations in multiple places while skimming through other PRs for both `reference` and `dispatcher`. My understanding was these were neccesarry because of this change, and these were just helper functions which called `dispatcher` and `reference` in runtime, but I haven't looked too closely at it.

packages/java-edition/src/mcfunction/index.ts
MulverineX: I think the way Misode is doing it here is fine

packages/java-edition/src/json/checker/index.ts
MulverineX: I agree with Neun that those helper methods were nice

PR #1246, opened 2024-06-09
✨ Add mcdoc validation for NBT

### LINKED ISSUE #1210: Add mcdoc validation for NBT
~~blocked by #1187~~

TheAfroOfDoom:
Closes #1210


packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -1188,12 +1193,24 @@ function getValueType(
 	}
 }
 
-export function getDefaultErrorReporter<T>(
+function getErrorRange<T extends AstNode>(
+	node: RuntimeNode<T>,
+	error: McdocCheckerError<T>['kind'],
+): Range {
```
misode, line 1199: I was confused why the build was failing, but this explains it. You're assuming here that `T` will always be an `AstNode`. The runtime checker does not assume that (as is demonstrated by the runtime checker tests, which use simple JS objects). Therefore, a `reportError` options will still need to be required, but you can just simplify those to not check the node type. Reverting this latest commit is probably the easiest.

misode (changes requested):

packages/mcdoc/src/runtime/checker/index.ts
TheAfroOfDoom: this commit is incomplete, I only pushed it to be able to continue working on it on my laptop 😄

misode (approved):
Looks complete to me. I made a few fixes while I was testing it.

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -1469,9 +1470,24 @@ function getValueType(
 	}
 }
 
+export function getErrorRangeDefault<T extends AstNode>(
```
NeunEinser, line 1473: should be `getDefaultErrorRange` (see also `getDefaultErrorReporter`)

NeunEinser (commented):
Nice work, looks good!

PR #1248, opened 2024-06-09
♻️ Rework mcdoc runtime checker tree structure

NeunEinser:


packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -256,140 +258,268 @@ export function isAssignable(
...
-	condensedChildErrors?: McdocCheckerError<T>[][]
 }
-interface CheckerTreeEntryNode<T> {
+
+interface CheckerTreeDefinitionNode<T> {
 	parent: CheckerTreeDefinitionNode<T> | undefined
-	runtimeKey?: RuntimeNode<T>
-	possibleRuntimeValues: CheckerTreeRuntimeNode<T>[]
-	typeDef: McdocType | undefined
+	entryNode: CheckerTreeNode<T>
```
misode, line 283: This field seems to be unused

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -256,140 +258,268 @@ export function isAssignable(
...
-				})
+		const encounterdErrors: CheckerTreeError<T>[] = []
 
-				if (lastValue !== oldParent) {
-					// Wait for all siblings to be evaluated first
-					break
+		for (const value of node.possibleValues) {
+			const inferredSimplified = simplify(
+				value.node.inferredType,
+				value.entryNode.runtimeKey,
```
misode, line 323: This can be replaced with `node.runtimeKey` I believe

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -256,140 +258,268 @@ export function isAssignable(
...
-						v =>
-							v.condensedChildErrors &&
-								v.condensedChildErrors.length > depth
-								? v.condensedChildErrors[depth]
-								: [],
-					),
-				}))
+				ans.possibleValues = (Array.isArray(c) ? c : c.possibleValues)
+					.map(v => ({
+						parent: value,
```
misode, line 342: This field can be removed

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -866,20 +933,47 @@ export function getPossibleTypes(
...
+	 * This key facilitates the %key lookup. This key should always be a member of
+	 * {@link value} and facilitates the `%key` lookup. This is the key of the member
+	 * the dispatcher originates from, while {@link value} is the struct containing
+	 * this key with a dispatcher associated with it.
+	 */
+	key: RuntimeNode<T> | undefined
+}
+/**
+ * Basically the same as a {@link CheckerTreeRuntimeNode}, just with a few fields
+ * that aren't needed here removed for simplicity.
```
misode, line 948: Not sure I'm a fan of introducing an entirely new set of types just because you don't need `children` here. That seems more confusing than reusing the `CheckerTreeRuntimeNode`. This rework makes the `simplify` function a bit more messy at first glance.

packages/mcdoc/src/runtime/checker/index.ts
NeunEinser: Yes, it can. Vice versa the line below could be `value.entryNode.parent`, but I agree it makes more sense to change this one.

packages/mcdoc/src/runtime/checker/index.ts
NeunEinser: I think I like keeping a reference back to the effective owner of this node, even though it is actually indeed unused in this case.

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -866,20 +933,47 @@ export function getPossibleTypes(
...
 
+interface SimplifyNode<T> {
+	value: SimplifyValueNode<T> | undefined
+	/**
+	 * This key facilitates the %key lookup. This key should always be a member of
+	 * {@link value} and facilitates the `%key` lookup. This is the key of the member
+	 * the dispatcher originates from, while {@link value} is the struct containing
+	 * this key with a dispatcher associated with it.
+	 */
+	key: RuntimeNode<T> | undefined
```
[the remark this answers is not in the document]
NeunEinser: Always.

The comment above attempts to explain this.

```rs
struct {
    foo: struct {
        id: string,
        data: minecraft:dispatcher[[id]]
    }
}
```
In this case, we want to do a lookup of id in the outer struct. To faciliate this we will have
```js
{
    value: { id: "bla", data: {...}, entryNode: { key: 'foo' } }
    key: 'data'
}
```
`[%key]` refers to `data` here, not `foo`.
`[[id]]` refers to `foo.id`, not `foo.data.id`.

So essentially, `value` is the value _containing_ `key`, whereas value.entryNode.key is the key the value itself is referred by.

packages/mcdoc/src/runtime/checker/index.ts
NeunEinser: The reason I do this is because it would be really annoying to need to construct a checker tree node with a bunch of empty arrays in e.g. line 1069, or require them to be set by another consumer.

misode: Ah, I missed that you also contruct these types from scratch. Makes sense to not reuse the types then.

packages/mcdoc/src/runtime/checker/index.ts
NeunEinser: (this changed to `runtimeNode` now with a similar semantic, due to validDefinitions being moved to CheckerTreeRuntimeNode. Probably still unused though, so it could be removed if you prefer that still)

NeunEinser:
I fixed the dispatcher issues and also added some tests for the case that was broken previously

__snapshots__/packages/mcdoc/test-out/runtime/checker.spec.js
```diff
@@ -617,6 +653,223 @@ exports['mcdoc runtime checker typeDefinition “struct { id: string, ...struct
...
+    },
+    "key": {
+      "kind": "literal",
+      "value": {
+        "kind": "string",
+        "value": "config"
+      }
+    }
+  }
+]
```
NeunEinser, line 695: This one is still wrong, not sure it works correctly on main though.

Expected would be a `some_missing_keys` error, since we aren't sure if `baz` or `config` is missing.

misode: It seems to be an issue caused by the error condenser, where it incorrectly sets `onlyCommonMissing` to "missing key config"

NeunEinser: Fixed.

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -256,140 +258,278 @@ export function isAssignable(
...
+	children: CheckerTreeNode<T>[]
 	node: RuntimeNode<T>
-	entryNode: CheckerTreeEntryNode<T>
+
+	condensedErrors: CheckerTreeError<T>[][]
 	validDefinitions: CheckerTreeDefinitionNode<T>[]
-	/// Each outer entry represents a layer of siblings
-	/// Each inner entry represents a specific condensed error within that layer.
-	condensedChildErrors?: McdocCheckerError<T>[][]
 }
```
misode, line 280: I would definitely keep these explanatory comments

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -564,298 +705,303 @@ function condenseErrorsAndFilterSiblings<T>(
...
+				.concat(
+					validDefinitions.flatMap(d => d.errors).filter(e =>
+						e.kind === 'some_missing_keys'
+					).map(e => e.node as RuntimeNode<T>),
+				)
+				.filter((v, i, arr) =>
+					arr.findIndex(o => o.originalNode === v.originalNode) === i
+				)
+				.map(n => ({
+					kind: 'some_missing_keys' as 'some_missing_keys',
```
misode, line 728: A good way to prevent some "unsafe" casts is by using `as const` on some of these objects

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -564,298 +705,303 @@ function condenseErrorsAndFilterSiblings<T>(
...
+	errors: McdocCheckerError<T>[]
+	childDefinitions: (McdocType | undefined)[]
+}
+
+function checkShallowly<T>(
+	runtimeNode: RuntimeNode<T>,
+	simplifiedInferred: SimplifiedMcdocTypeNoUnion,
+	children: RuntimeUnion<T>[],
+	typeDef: SimplifiedMcdocTypeNoUnion,
 	options: McdocCheckerOptions<T>,
```
misode, line 752: I very much like that you made this function not use any of the graph types, and also got rid of the mutations/side effects!

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -564,298 +705,303 @@ function condenseErrorsAndFilterSiblings<T>(
...
+				typeDef.value.value !== simplifiedInferred.value.value)) ||
+		// TODO handle enum field attributes
+		(typeDef.kind === 'enum' &&
+			(simplifiedInferred.kind !== 'literal' ||
+				!typeDef.values.some(v =>
+					v.value === simplifiedInferred.value.value
+				)))
+	) {
+		return {
+			childDefinitions: Array(children.length).fill(undefined),
```
misode, line 772: What is the point of this? Since there is a type mismatch, why not return an empty list of definitions?

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -564,298 +705,303 @@ function condenseErrorsAndFilterSiblings<T>(
...
-							),
-						)
+			for (let i = 0; i < children.length; i++) {
+				const child = children[i]
+				if (Array.isArray(child)) {
+					continue
+				}
+				if (
+					child.key.inferredType.kind === 'literal' &&
+					child.key.inferredType.value.kind === 'string'
```
misode, line 821: Maybe it's time to add some helpers for these like `isStringLiteral(type: McdocType): type is { ... }`?

PR #1249, opened 2024-06-09
⚡️ Optimize struct field duplicate filtering

misode:
Replaces the old `O(n^2)` field de-duplication check in the struct simplify case. Now a map of fields is used for literal keys, and only the complex keys are checks if they are assignable.

### DIFF
--- a/packages/mcdoc/src/runtime/checker/index.ts
+++ b/packages/mcdoc/src/runtime/checker/index.ts
@@ -1111,7 +1111,9 @@ export function simplify<T>(
 				} else if (key.kind === 'literal' && key.value.kind === 'string') {
 					literalFields.set(key.value.value, field)
 				} else if (key.kind === 'union') {
-					key.members.forEach(m => addField(m, field))
+					key.members.forEach(m =>
+						addField(m, { ...field, optional: true })
+					)
 				} else {
 					complexFields = complexFields.filter(other =>
 						!isAssignable(
--- a/packages/mcdoc/src/runtime/checker/index.ts
+++ b/packages/mcdoc/src/runtime/checker/index.ts
@@ -1099,7 +1099,36 @@ export function simplify<T>(
 			}
 			return { kind: 'union', members: members }
 		case 'struct':
-			const fields: SimplifiedStructTypePairField[] = []
+			const literalFields = new Map<string, StructTypePairField>()
+			let complexFields: SimplifiedStructTypePairField[] = []
+
+			function addField(
+				key: string | SimplifiedMcdocType,
+				field: StructTypePairField,
+			) {
+				if (typeof key === 'string') {
+					literalFields.set(key, field)
+				} else if (key.kind === 'literal' && key.value.kind === 'string') {
+					literalFields.set(key.value.value, field)
+				} else if (key.kind === 'union') {
+					key.members.forEach(m => addField(m, field))
+				} else {
+					complexFields = complexFields.filter(other =>
+						!isAssignable(
+							typeof other.key === 'string'
+								? {
+									kind: 'literal',
+									value: { kind: 'string', value: other.key },
+								}
+								: other.key,
+							key,
+							options.context,
+							options.isEquivalent,
+						)
+					)
+					complexFields.push({ ...field, key })
+				}
+			}
 			for (const field of typeDef.fields) {
 				if (field.kind === 'pair') {
 					// Don't simplify the value here. We need to have the correct `node` and `parents`, which we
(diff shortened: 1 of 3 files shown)


packages/mcdoc/test/runtime/checker.spec.ts
```diff
@@ -11,7 +11,7 @@ import { describe, it } from 'mocha'
 import snapshot from 'snap-shot-it'
 import { TextDocument } from 'vscode-languageserver-textdocument'
 
-describe('mcdoc runtime checker', () => {
+describe.only('mcdoc runtime checker', () => {
```
TheAfroOfDoom, line 14: woops

```suggestion
describe('mcdoc runtime checker', () => {
```

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -1099,7 +1099,36 @@ export function simplify<T>(
 			}
 			return { kind: 'union', members: members }
 		case 'struct':
-			const fields: SimplifiedStructTypePairField[] = []
+			const literalFields = new Map<string, StructTypePairField>()
+			let complexFields: SimplifiedStructTypePairField[] = []
```
TheAfroOfDoom, line 1103: these collections live inside `simplify` and not e.g. globally, guessing this is intentional. does this mean that these collections only apply to the struct at the current level, and we won't re-use keys for structs on typeDefs above/below the current struct?

(my wording my be a bit off here)

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -1108,43 +1137,58 @@ export function simplify<T>(
...
+				fields: [
+					...complexFields,
+					...[...literalFields.entries()].map(([key, field]) =>
+						({
+							...field,
+							key: {
+								kind: 'literal',
+								value: { kind: 'string', value: key },
+							},
+						}) satisfies SimplifiedStructTypePairField
```
TheAfroOfDoom, line 1189: :O ive never seen `satisfies` before

packages/mcdoc/test/runtime/checker.spec.ts
misode: Damn this always happens 🤦

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -1099,7 +1099,36 @@ export function simplify<T>(
...
+			function addField(
+				key: string | SimplifiedMcdocType,
+				field: StructTypePairField,
+			) {
+				if (typeof key === 'string') {
+					literalFields.set(key, field)
+				} else if (key.kind === 'literal' && key.value.kind === 'string') {
+					literalFields.set(key.value.value, field)
+				} else if (key.kind === 'union') {
+					key.members.forEach(m => addField(m, field))
```
NeunEinser, line 1114: Mark field as optional explicitly?

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -1108,43 +1137,58 @@ export function simplify<T>(
...
-							of.key,
-							f.key,
-							options.context,
-							options.isEquivalent,
-						)
-					) === i
-				),
+				fields: [
+					...complexFields,
+					...[...literalFields.entries()].map(([key, field]) =>
```
TheAfroOfDoom, line 1182: why do we flatten the `Map` we just made instead of returning it as-is? like

```suggestion
				fields: {
					complexFields,
					literalFields,
				}
```

and re-using it later...?

by flattening it, it looks like we're just gonna have to redo the `addField` stuff again later

packages/mcdoc/src/runtime/checker/index.ts
misode: Correct. These are temporary collecting variables only used in the current struct case, and at the end combines into a result.

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -1099,7 +1099,36 @@ export function simplify<T>(
...
+				field: StructTypePairField,
+			) {
+				if (typeof key === 'string') {
+					literalFields.set(key, field)
+				} else if (key.kind === 'literal' && key.value.kind === 'string') {
+					literalFields.set(key.value.value, field)
+				} else if (key.kind === 'union') {
+					key.members.forEach(m => addField(m, field))
+				} else {
+					complexFields = complexFields.filter(other =>
```
NeunEinser, line 1116: Not sure if intentional, but this will keep fields whose keys are assignable to previous types if the previous type is broader.

For example, this would not be deduplicated:
```rs
struct {
    [string]: int
    [#[id=block] string]: int
}
```

This is because you are only checking if `string` is assignable to `#[id=block] string` (which it won't be when attributes are implemented), but not the other way around.

I think if we want to allow for this, it would make more sense to check the other way around and allow this:

```rs
struct {
    [string]: int
    [#[id=block] string]: int
}
```

packages/mcdoc/src/runtime/checker/index.ts
misode: Because some of the complex fields are not represented in a map. It is true that nested struct spreading will cause the literal fields to be converted between a map and a list a few times. Maybe this should be improved. But simply returning a map here is not possible.

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -1108,43 +1137,58 @@ export function simplify<T>(
...
+						}
+					} else {
+						const type = McdocType.toString(simplifiedSpreadType)
+						options.context.logger.warn(
+							`Tried to spread non-struct type ${type}`,
+						)
 					}
 				}
 			}
+			complexFields = complexFields.filter(other =>
```
NeunEinser, line 1159: Not sure we really need this one. I wouldn't expect a complex type to be assignable to a literal.

packages/mcdoc/src/runtime/checker/index.ts
misode: Sorry, what do you mean with this?

NeunEinser: We said we want any non-literal struct member to be implicitly optional.

Here, you are flattening a union. Since a union is non-literal, this field should be implicitly optional.

If this is a union of literals, there is no way to know this later on though, so if you flatten the union here, you should mark all it's members optional explicitly.

packages/mcdoc/src/runtime/checker/index.ts
misode: You're right, that way around makes more sense 👍

packages/mcdoc/src/runtime/checker/index.ts
misode: Fair. I will take it out but leave a comment in case we come across situations where we need it.

packages/mcdoc/src/runtime/checker/index.ts
misode: I got rid of it 👀

NeunEinser (approved):
Nice work!

packages/mcdoc/src/runtime/checker/index.ts
NeunEinser: Yeah, this would need some more rework to facilitate this, and the performance issue we ran into is not really related to this.

PR #1250, opened 2024-06-09
🎨 Change format rules

NeunEinser:
Everything up for discussion

- Added `object-shorthand` eslint rule
- Changed `operatorPosition` to `nextLine`
- Changed line width from 80 to 100
- Reformatted project using `preferSingleLine: true` temporarily, to illustrate how it might look like with the new line length (this rule is experimental, makes formatting significantly slower, and crashes when trying to format `java-edition/src/mcfunction/tree/patch.ts`, so it definitley shouldn't be enabled by default, and I like the idea of leaving the option to the author to potentially wrap lines more than neccessary)

This is meant as an illustration of how the formatting could be improved. I think bumping the line width a bit like this is a good compromise, I tried `120` initially, but that seemed a bit too much in some cases.

### DIFF
--- a/packages/core/src/common/externals/NodeJsExternals.ts
+++ b/packages/core/src/common/externals/NodeJsExternals.ts
@@ -27,27 +27,15 @@ const gunzip = promisify(zlib.gunzip)
 const gzip = promisify(zlib.gzip)
 
 class NodeJsExternalDownloader implements ExternalDownloader {
-	get(
-		uri: RemoteUriString,
-		options: ExternalDownloaderOptions = {},
-	): Promise<Uint8Array> {
+	get(uri: RemoteUriString, options: ExternalDownloaderOptions = {}): Promise<Uint8Array> {
 		const protocol = new Uri(uri).protocol
 		return new Promise((resolve, reject) => {
 			const backend = protocol === 'http:' ? http : https
 			backend.get(uri, options, (res: IncomingMessage) => {
 				if (res.statusCode !== 200) {
-					reject(
-						new Error(
-							`Status code ${res.statusCode}: ${res.statusMessage}`,
-						),
-					)
+					reject(new Error(`Status code ${res.statusCode}: ${res.statusMessage}`))
 				} else {
-					resolve(
-						promisifyAsyncIterable(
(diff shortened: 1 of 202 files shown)


packages/core/src/common/externals/NodeJsExternals.ts
```diff
@@ -169,12 +152,13 @@ class ChokidarWatcherWrapper extends EventEmitter implements FsWatcher {
...
-			.on('change', (path) => this.emit('change', uriFromPath(path)))
-			.on('unlink', (path) => this.emit('unlink', uriFromPath(path)))
-			.on('error', (e) => this.emit('error', e))
+		this.#watcher = watcher.on('ready', () => this.emit('ready')).on(
+			'add',
+			(path) => this.emit('add', uriFromPath(path)),
+		).on('change', (path) => this.emit('change', uriFromPath(path))).on(
+			'unlink',
+			(path) => this.emit('unlink', uriFromPath(path)),
+		).on('error', (e) => this.emit('error', e))
```
SPGoding, line 161: `preferSingleLine: true` definitely ruined this lol but in general I'm fine with these changes

SPGoding (approved):

packages/core/src/common/externals/NodeJsExternals.ts
NeunEinser: _I can change it to 99 lmao_

NeunEinser: I fixed the formatting of this file manually.

TheAfroOfDoom: 99 feels weird, id be happier with 90 or even 95 as a rounder number. but /shrug

misode (approved):
Looks pretty good. There are some places where the formatter has made it a bit worse, but I'm sure we can cope with that. I wouldn't increase the column limit anymore, but I agree that 80 is maybe a bit too conservative.

Calverin:
I think the length change is a big improvement! I've noticed some areas where the length cutoff seemed a bit much so I'll be happy to see this implemented.

PR #1252, opened 2024-06-11
♻️ Condense mcdoc runtime errors in a unified way

NeunEinser:
This should fix all bugs with error condensing and valid filtering definitions.


packages/mcdoc/src/runtime/checker/error.ts
```diff
@@ -0,0 +1,463 @@
...
+	definition: CheckerTreeDefinitionNode<T>
+	errors: McdocRuntimeError<T>[]
+}
+
+export function condenseErrorsAndFilterSiblings<T>(
+	definitions: {
+		definition: CheckerTreeDefinitionNode<T>
+		errors: McdocRuntimeError<T>[]
+	}[],
+	context: McdocCheckerContext<T>,
```
misode, line 114: Context is never used throughout this whole function and `condense`, but I assume you're keeping it in potentially for future use cases, and maybe logging?

packages/mcdoc/src/runtime/checker/error.ts
```diff
@@ -0,0 +1,463 @@
...
+export interface ErrorCondensingDefinition<T> {
+	definition: CheckerTreeDefinitionNode<T>
+	errors: McdocRuntimeError<T>[]
+}
+
+export function condenseErrorsAndFilterSiblings<T>(
+	definitions: {
+		definition: CheckerTreeDefinitionNode<T>
+		errors: McdocRuntimeError<T>[]
+	}[],
```
misode, line 113: This should be `definitions: ErrorCondensingDefinition<T>[]`

packages/mcdoc/src/runtime/checker/error.ts
NeunEinser: Ah, this was needed for `isAssignable` before I replaced it with `equals`

packages/mcdoc/src/runtime/checker/error.ts
NeunEinser: Yeah I think this was nuked by a merge conflict

packages/mcdoc/src/runtime/checker/error.ts
```diff
@@ -0,0 +1,463 @@
...
+		condensedErrors: errors,
+	}
+}
+
+function condense<T, E extends McdocRuntimeError<T>>(
+	validDefinitions: ErrorCondensingDefinition<T>[],
+	context: McdocCheckerContext<T>,
+	is: (e: McdocRuntimeError<T>) => e is E,
+	equals: (a: E, b: E) => boolean,
+	combineAlternatives?: (errors: E[]) => E,
```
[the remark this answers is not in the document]
NeunEinser: Isn't this the same (plus two curly braces)?

packages/mcdoc/src/runtime/checker/error.ts
```diff
@@ -0,0 +1,463 @@
...
+			continue
+		}
+		definitions.push(
+			...group
+				.filter(v =>
+					!definitionGroups.some((og, oi) =>
+						(oi > i || og.length === 1)
+							&& predicate
+							? og.some(ov => predicate(v, ov))
+							: og.includes(v)
```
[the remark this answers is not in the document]
NeunEinser: This actually should be improved a bit over what we had before, as the old condensing needed to do similar things, but just did a flatMap and had a "truer" O(n^2), whereas here it's pretty optimized and only checks things it actually needs to, and only when conflicting errors actually exist (if it's length one it takes the short circuit)

If we run into issues here, we probably have to make some sacrifices and accept some errors to contain duplicated messages, but I think this one should be pretty fine.

packages/mcdoc/src/runtime/checker/error.ts
NeunEinser: Ah yeah good catch, this broke when I copied stuff over again after merging `internal` error. :upside_down_face:

packages/mcdoc/src/runtime/checker/error.ts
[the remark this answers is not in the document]
misode: The lambda functions are named when you use this function

misode:
Maybe it wouldn't hurt adding some tests for the more complicated examples where this PR improves the error condensing?

NeunEinser:
> Maybe it wouldn't hurt adding some tests for the more complicated examples where this PR improves the error condensing?

Yeah I agree, I think I would like to do them separetely though before I need to fix more merge conflicts :D

Not sure I will have time before the weekend. ;)

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -1,7 +1,8 @@
-import * as core from '@spyglassmc/core'
-import { arrayToMessage, localeQuote, localize } from '@spyglassmc/locales'
+import { Range, Source } from '@spyglassmc/core'
+import type { CheckerContext, FullResourceLocation } from '@spyglassmc/core'
```
misode, line 2: I prefer having the `core` prefix for these, as is this case with most other files

NeunEinser: ah well, I usually just do that when there are lots of imports from one package, but I can change it later ¯\\\_(ツ)\_/¯

misode (approved):

PR #1255, opened 2024-06-12
✨ Add nbt path mcdoc checking

### LINKED ISSUE #1227: Add mcdoc based nbt path checking
Relates to #1210

misode:
Closes #1227


MulverineX (approved):

packages/nbt/src/checker/index.ts
```diff
@@ -286,18 +279,149 @@ export function blockStates(
...
+					}
+					// Never reachable
+					return []
+				},
+				reportError: (error) => {
+					if (
+						error.kind === 'missing_key' ||
+						error.kind === 'invalid_collection_length'
+					) {
+						return
```
TheAfroOfDoom, line 384: why is this an empty return?

misode: Due to the way nbt paths are represented to the mcdoc checker (see `NbtPathLink`), the checker will report lots of "missing key" errors, because you're only visiting one branch of the children. Likewise when you're indexing into a list, there is no way for you to specify all N children.

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -1425,12 +1428,18 @@ export function simplify<T>(
...
-						ctx.logger.warn(
-							`Tried to spread non-struct type ${type}`,
-						)
+					} else if (simplifiedSpreadType.kind === 'union') {
+						for (const member of simplifiedSpreadType.members) {
+							if (member.kind === 'struct') {
+								for (const field of member.fields) {
+									// TODO: technically, if a field is required in all members of
+									// this union, it could be added as a required field
+									addField(field.key, { ...field, optional: true })
```
NeunEinser, line 1437: This is a bit problematic as it really should be resolved as another union.

E.g. if you have sth like this
```rs
struct {
    ...( struct { foo: string } | struct { bar: string } ),
    baz?: boolean
}
```

The way you are doing it here, both `foo` and `bar` will be optional so an empty struct would be valid and not raise an error, while the expected behavior would be that it reports `missing one of the keys "foo" or "bar"`.

So if we want to handle to handle unions here, the only way is to resolve them like this in the above example:
```rs
(
    struct { foo: string, baz?: bool }
    | struct { bar: string, baz?: bool }
)
```

This would also automatically take care of the TODO you provided here.

packages/nbt/src/checker/index.ts
NeunEinser: Ideally for `some_list[5]` you would create 4 `unsafe` types as children (or empty union?) and report a custom error when the maximum is exceeded.

packages/nbt/src/checker/index.ts
```diff
@@ -286,18 +279,149 @@ export function blockStates(
...
+				stringAttacher: (node, attacher) => {
+					if (!core.StringNode.is(node)) return
+					attacher(node)
+					if (node.children) {
+						core.AstNode.setParents(node)
+						// Because the runtime checker happens after binding, we need to manually call this
+						core.binder.dispatchSync(node, ctx)
+						core.checker.dispatchSync(node, ctx)
+					}
+				},
```
NeunEinser, line 411: Would it make sense to add a general getDefaultTypeInfoAttacher and getDefaultStringAttacher? The code here looks pretty similar across different consumers.

packages/mcdoc/src/runtime/checker/index.ts
misode: Hmm you're right that would be an issue. Do we have any such cases in vanilla-mcdoc? I worry about exploding the size of the returned "simplified" type

packages/nbt/src/checker/index.ts
misode: Agreed, this could use a DRY-refactoring

packages/mcdoc/src/runtime/checker/index.ts
misode: The only cases where imo we want to spread a union of structs is for entity, block, and item NBT. And in all those cases all fields are optional.

NeunEinser: This will as implemented right now be used for any dispatcher fallback case, for example when no known id is supplied somewhere.

There are definitely dispatched values that have required fields.

misode: That's why I think for most dispatcher fallbacks (except for the above 3 cases), we should not return a union of all. If the type is not available we should return an empty union.

NeunEinser: I'm not sure on this. The main reason it is done differently is that the [mcdoc reference](https://spyglassmc.com/user/mcdoc/#indexing-on-a-type) says the default value should be fallback.

> The default value used for all cases (including the two mutable special keys, `%none` and `%unknown`) is the [fallback case](https://spyglassmc.com/user/mcdoc/#dispatcher-fallback-case).

PR #1276, opened 2024-06-14
✨ Run mcdoc checker in various command arguments

### LINKED ISSUE #1265: Run mcdoc checker for inlined loot resources

### LINKED ISSUE #1267: Dispatch entity, block and item NBT using the correct command context

### LINKED ISSUE #1271: Run mcdoc checker in `nbt=` selector argument

### LINKED ISSUE #1272: Run mcdoc checker as part of `block` command argument
For example in the command:
```hs
setblock ~ ~ ~ command_block{Command:"say hi"}
```

misode:
- [x] Close #1265
- [x] Close #1271
- [x] Close #1272
- [x] Close #1267


MulverineX (approved):

__snapshots__/packages/java-edition/test-out/mcfunction/parser/argument/minecraftNbtTag.spec.js
```diff
@@ -24,152 +33,197 @@ exports['mcfunction argument minecraft:nbt_tag Parse ""foo"" 1'] = {
...
     },
-    "value": "0"
+    "children": [
+      {
+        "type": "nbt:long",
+        "range": {
+          "start": 0,
+          "end": 2
+        },
+        "value": "0"
```
Calverin, line 111: Why is the value for a long stored as a string compared to doubles, bytes, and ints?

Calverin: Additionally it might be a good idea to have tests for 1.0f, 1.0d and maybe a negative number but it's not too important

SPGoding: JavaScript uses 64 bit IEEE floating point numbers which does not have enough precision to store a 64 bit long integer, so an arbitrary-precision integer type `BitInt` is used instead, which gets serialized to strings in our snapshot implementation.

Calverin: Okay got it, just wanted to make sure it wasn't a mistake :+1:

PR #1282, opened 2024-06-15
🐛 Fix `Reset Project Cache` command

### LINKED ISSUE #1258: Ensure the `Reset Project Cache` command actually works

misode:
- Fixes #1258

The issue was caused because the `vanilla-datapack` and `vanilla-mcdoc` dependencies relied on the downloader setting the `out.cacheUri` field. This did not happen when memory cache was found (because only the raw buffer was stored in the memory cache).

The memory cache is only intended to last 15 seconds, but due to a logic error, memory cache entries never expired.

### DIFF
--- a/packages/core/src/service/Downloader.ts
+++ b/packages/core/src/service/Downloader.ts
@@ -207,5 +207,8 @@ interface Job<R> {
 	}
 	transformer: (data: Uint8Array) => PromiseLike<R> | R
 	options?: ExternalDownloaderOptions
+	/**
+	 * If set, caches the result in memory. Time in milliseconds.
+	 */
 	ttl?: number
 }
--- a/packages/core/src/service/Downloader.ts
+++ b/packages/core/src/service/Downloader.ts
@@ -13,11 +13,15 @@ export interface DownloaderDownloadOut {
 	checksum?: string
 }
 
+interface MemoryCacheEntry {
+	buffer: Uint8Array
+	time: number
+	cacheUri?: string
+	checksum?: string
+}
+
 export class Downloader {
-	readonly #memoryCache = new Map<
-		string,
-		{ buffer: Uint8Array; time: number }
-	>()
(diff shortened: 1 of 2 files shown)


packages/core/src/service/Downloader.ts
```diff
@@ -31,11 +35,14 @@ export class Downloader {
 	): Promise<R | undefined> {
 		const { id, cache, uri, options, transformer, ttl } = job
 		if (ttl && this.#memoryCache.has(uri)) {
-			const { buffer, time } = this.#memoryCache.get(uri)!
-			if (time <= performance.now() + ttl) {
+			const memoryCacheEntry = this.#memoryCache.get(uri)!
```
TheAfroOfDoom, line 38: annoying that we need the non-null assertion here. I see that it can't be undefined because of the `has` above

TheAfroOfDoom (approved):
so this bug is why we were always recommending we delete the cache manually? with `Open Cache Folder`

packages/core/src/service/Downloader.ts
misode: Yeah bit of a typescript quirk, but not that important. We have this pattern in a couple of other places.

misode:
> so this bug is why we were always recommending we delete the cache manually? with `Open Cache Folder`

Yes, hopefully that won't be necessary anymore. We'll have to monitor when people get symbol cache issues and check if the command works fine for them.

packages/core/src/service/Downloader.ts
```diff
@@ -31,11 +35,14 @@ export class Downloader {
...
+			const { buffer, time, cacheUri, checksum } = memoryCacheEntry
+			if (performance.now() <= time + ttl) {
```
SPGoding, line 40: What was I smoking lol

SPGoding (approved):

PR #1283, opened 2024-06-15
✨ Improve command argument mcdoc attributes

### LINKED ISSUE #1260: Handle mcdoc `command_argument` attributes with properties
Some argument parsers require properties in the command tree, right now only the parser ID is passed, which makes some arguments impossible to use in mcdoc.

Affected parsers:
* `score_holder`
* `entity`
* ???

misode:
- Closes #1260


misode:
I'm happy with this PR now.

packages/java-edition/src/mcfunction/mcdocAttributes.ts
```diff
@@ -0,0 +1,85 @@
...
+	rootTreeNode: mcf.RootTreeNode,
+) {
+	mcdoc.runtime.registerAttribute(meta, 'command', () => undefined, {
+		// TODO: validate slash
+		// TODO: fix completer inside commands
+		stringParser: () => {
+			return mcf.command(rootTreeNode, parser.argument)
+		},
+	})
+	mcdoc.runtime.registerAttribute(meta, 'text_component', () => undefined, {
```
NeunEinser, line 24: How are inline SNBT handled in e.g. JSON strings? I could see some kind of inline reference here coupled with the normal JSON parser instead of hardcoding text_component specifically here

packages/java-edition/src/mcfunction/mcdocAttributes.ts
```diff
@@ -0,0 +1,85 @@
...
+		stringParser: () =>
+			makeInfallible(
+				parser.jsonParser('::java::server::util::text::Text'),
+				localize('text-component'),
+			),
+	})
+	mcdoc.runtime.registerAttribute(meta, 'objective', () => undefined, {
```
NeunEinser, line 31: I thought we implemented those using command arguments (objectuve, team, score_holder, selector and item_slot)? Was there an issue with this? Shouldn't it be possible to re-use the general mock function from mcfunction here as well?

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -503,16 +503,34 @@ export function typeDefinition<T>(
...
+									localize('mcdoc.runtime.checker.value'),
+								),
+								core.Range.create(start, src.skipRemaining()),
+							)
+							return
+						} else if (src.canRead()) {
+							options.context.err.report(
+								localize('mcdoc.runtime.checker.trailing'),
+								core.Range.create(src.cursor, src.skipRemaining()),
+							)
```
NeunEinser, line 529: error reproting should really be restricted to `reportErrors`, and they should be thrown when the nodes and definitions are first encountered for error condensing and filtering definitions.

packages/java-edition/src/mcfunction/mcdocAttributes.ts
misode: * Some command arguments require properties. Having them as separate attributes makes it easier to validate
* Shorter attributes makes the mcdoc source more readable
* I refactored the handler method to accept a parser, so they can be registered very compact here

packages/mcdoc/src/runtime/checker/index.ts
misode: I disagree with that, see also my replies on #1245. If attributes want to affect the definition filtering they need to use `checkInferred` (like `id` does), The `stringParser` handler should assume that this is the only typedef possible for the node and runs after type info has been attached.

packages/java-edition/src/mcfunction/mcdocAttributes.ts
misode: We have a `nbt` attribute that takes a mcdoc type, so yeah we could replace `text_component` with a `json` attribute doing something similar.

packages/mcdoc/src/runtime/checker/index.ts
NeunEinser: Well, the point is, there is no reason to assume there is only one possible union type that has a string parser. Maybe this is true right now, but I still don't think this is the correct approach down the line. It might be fine for 4.0.

packages/java-edition/src/mcfunction/mcdocAttributes.ts
NeunEinser: > * I refactored the handler method to accept a parser, so they can be registered very compact here

I would say, it is still some overhead (and more cases may occur in the future), and it would be really nice to have a command_argument attribute anyway for custom mcdoc, especially considering macros.

PR #1286, opened 2024-06-15
⚡ Optimize `isAssignable` for literal types

misode:
The vast majority (if not all) calls to `isAssignable` compare two literal types. This PR introduces a fast path for those cases, assuming that two literal types of the same kind are always equal to each other when their values match.

A few benchmarks:
* 9.39ms -> 2.36ms
* 170.11ms -> 117.26ms
* 320.82ms -> 274.17ms

That last one in particular was
```
data merge entity @s {Air:1b,state:"rolling",Passengers:[{Air:0}]}
```

Additionally this PR removes some of the attribute warnings, bringing the total time of that last example all the way down to 8.08ms!

### DIFF
--- a/packages/mcdoc/src/runtime/checker/index.ts
+++ b/packages/mcdoc/src/runtime/checker/index.ts
@@ -165,6 +165,12 @@ export function isAssignable(
 	ctx: core.CheckerContext,
 	isEquivalent?: NodeEquivalenceChecker,
 ): boolean {
+	if (
+		assignValue.kind === 'literal' && typeDef.kind === 'literal' &&
+		assignValue.value.kind === typeDef.value.kind
+	) {
+		return assignValue.value.value === typeDef.value.value
+	}
 	let ans = true
 	const options: McdocCheckerOptions<McdocType> = {
 		context: ctx,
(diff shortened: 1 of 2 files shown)


misode:
These are too good not to merge, sorry. If this literal isAssignable check turns out to be wrong, blame me later.

NeunEinser:
> Additionally this PR removes some of the attribute warnings, bringing the total time of that last example all the way down to 8.08ms!

Would this be smaller when all attributes are implemented? I think I prefer having warnings here.

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -165,6 +165,13 @@ export function isAssignable(
 	ctx: core.CheckerContext,
 	isEquivalent?: NodeEquivalenceChecker,
 ): boolean {
+	if (
+		assignValue.kind === 'literal' && typeDef.kind === 'literal' &&
+		assignValue.value.kind === typeDef.value.kind &&
+		!assignValue.attributes && !typeDef.attributes
+	) {
+		return assignValue.value.value === typeDef.value.value
```
NeunEinser, line 173: Maybe if this is false, only return if both kinds do not have any attributes set? Or alternatively, call the appropiate mapping attribute function here as well?

misode:
> Would this be smaller when all attributes are implemented? I think I prefer having warnings here.

There were a lot of duplicate warnings, all in the hot simplify path. If we want warnings for unimplemented attributes, I think it would be better to do this during the mcdoc binding process, and also de-duplicate them.

PR #1289, opened 2024-06-16
✨ Run mcdoc checker on new particle syntax

misode:


MulverineX (approved):

packages/java-edition/src/mcfunction/checker/index.ts
```diff
@@ -188,7 +189,31 @@ function nbtChecker(dispatchedBy?: core.AstNode): core.SyncChecker<NbtNode> {
...
 const particle: core.SyncChecker<ParticleNode> = (node, ctx) => {
-	core.checker.dispatchSync(node, ctx)
+	const id = core.ResourceLocationNode.toString(node.id, 'short')
+	const release = ctx.project['loadedVersion'] as ReleaseVersion | undefined
+	if (release && ReleaseVersion.cmp(release, '1.20.5') < 0) {
+		return
+	}
+	const options = node.children?.find(nbt.NbtCompoundNode.is)
+	ctx.symbols.query(ctx.doc, 'mcdoc/dispatcher', 'minecraft:particle', id)
+		.ifDeclared(() => {
```
misode, line 199: This won't handle versions changes (let's say mojang makes an existing particle require options), so maybe a hardcoded list here is a better idea. I don't fancy running the mcdoc simplifying to filter out `#[since]` attributes and check whether the dispatched type is an empty union.

PR #1290, opened 2024-06-16
✨ Show doc comments on mcdoc hover

### LINKED ISSUE #1264: Show mcdoc doc comment on hover

TheAfroOfDoom:
- Fixes #1264 

**Important:**
When switching to this PR, run `Reset Project Cache` to allow the doc comments to be included in the symbol table

image


NeunEinser (approved):
In the future, this should use translations and the binder should provide a translation key for a doc comment instead (which could then also be used by mcdoc cli to generate translations).

But I think this is totally fine for 4.0, though maybe @MulverineX has an opinion about this.

misode (approved):

PR #1292, opened 2024-06-16
🐛 Allow nbt byte to be assignable to boolean

### LINKED ISSUE #1274: Fix issues with booleans in NBT
Make sure that bytes `0` and `1` are assignable to `boolean`

image

misode:
- Fix #1274 

Future work should allow `isEquivalent` to check the value of a literal

### DIFF
--- a/packages/nbt/src/checker/index.ts
+++ b/packages/nbt/src/checker/index.ts
@@ -65,6 +65,10 @@ export function typeDefinition(
 			{
 				context: ctx,
 				isEquivalent: (inferred, def) => {
+					if (def.kind === 'boolean') {
+						// TODO: this should check whether the value is 0 or 1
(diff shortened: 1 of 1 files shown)


NeunEinser:
> Future work should allow `isEquivalent` to check the value of a literal

Yeah, we should discuss how exactly we should handle this.

For most numeric types, I think I would prefer handling it in mcdoc runtime, e.g. if it's in principle assignable to a byte, check if the value is in the correct range.

We could do the same for boolean, but this is maybe a bit hacky, as a boolean cannot really be converted to a number. My hunch is, maybe a custom converter function could be used, which will be called when doing a non-trivial conversion needs to happen for the checker (which then could also return undefined when it fails), but I am not too sure about this yet.

NeunEinser (approved):

PR #1293, opened 2024-06-16
🐛 Attach all validDefinitions to nodes

### LINKED ISSUE #1275: Allow multiple type definitions to be attached to a node
Right now `attachTypeInfo` can only attach one `SimplifiedMcdocType` to a node, but there are lots of sitations where multiple `validDefinitions` are available. In those cases either a union, or a list of types should be attached.

Example: `data merge block ~ ~ ~ {}` - autocompletions here only show the fields for the `banner` block entity, because that just so happens to be the first in `validDefinition`. It should show all possible fields.

misode:
- Fixes #1275

### DIFF
--- a/packages/mcdoc/src/runtime/checker/index.ts
+++ b/packages/mcdoc/src/runtime/checker/index.ts
@@ -438,40 +438,40 @@ export function typeDefinition<T>(
 		}
 	}
 
-	// TODO: attach a combination of all possible definitions
-	const attached = new Set<T>()
-
 	function attachTypeInfo(node: CheckerTreeRuntimeNode<T>) {
-		for (const def of node.validDefinitions) {
-			if (!attached.has(node.node.originalNode)) {
-				options.attachTypeInfo(node.node.originalNode, def.typeDef)
-				const attributes = def.typeDef.attributes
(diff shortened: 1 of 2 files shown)


packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -438,40 +438,40 @@ export function typeDefinition<T>(
...
 				})
-				attached.add(node.node.originalNode)
-			}
+			})
+		} else if (node.validDefinitions.length > 1) {
+			options.attachTypeInfo(node.node.originalNode, {
+				kind: 'union',
+				members: node.validDefinitions.map(d => d.typeDef),
+			})
+			// TODO: don't ignore attributes with a stringParser?
```
NeunEinser, line 474: Regarding the todo, this case implies multiple possible definitions, so it shouldn't really be an issue, at least with how string attaching is handled right now.

There are probably cases where it happens, but it should be sth like `#[...] string | struct { ... }`, where as soon as you know it's a string the child can be added correctly, and if you don't it would be incorrect as we don't even know if it is a string.

NeunEinser (approved):

PR #1295, opened 2024-06-16
🐛 Escape completions in string

### LINKED ISSUE #1284: Completions in escaped strings are broken
The string completer should transform completion items, escaping the necessary characters.

image
image

misode:
- Fixes #1284 

I don't like the implementation of this yet, so this may need to be improved. Ideally they'd also share the implementation of the core string completer (which has a TODO for this issue)


NeunEinser:
This looks good to me, but I am not too familiar with completer code, so I don't really understand the issues you highlighted in the description.

PR #1311, opened 2024-06-17
🐛 Fix type info not being attached to keys by mcdoc runtime

### LINKED ISSUE #1263: Attach type info and run string attacher on keys
Right now the `stringAttacher` option is only called for values, not for keys. This is required for complex field keys with an `id` attribute. Example: `components` map in `set_components` loot function.

This is not super trivial to do, it looks like the checker graph is missing some information here, possibly as a result of #1248

NeunEinser:
Fixes #1263

Auto complete does not appear to work for this kind of key yet.

### DIFF
--- a/packages/mcdoc/src/runtime/checker/index.ts
+++ b/packages/mcdoc/src/runtime/checker/index.ts
@@ -450,62 +450,8 @@ export function typeDefinition<T>(
 		}
 	}
 
-	function attachTypeInfo(node: CheckerTreeRuntimeNode<T>) {
-		function handleStringAttachers(
-			runtimeValue: RuntimeNode<T>,
-			typeDef: SimplifiedMcdocTypeNoUnion,
-		) {
-			return handleAttributes(typeDef.attributes, options.context, (handler, config) => {
-				const parser = handler.stringParser?.(config, options.context)
-				if (!parser) {
-					return
-				}
-				options.stringAttacher(runtimeValue.originalNode, (node) => {
-					const src = new core.Source(node.value, node.valueMap)
-					const start = src.cursor
-					const child = parser(src, options.context)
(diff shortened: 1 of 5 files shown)


packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -445,15 +451,16 @@ export function typeDefinition<T>(
...
 
 	function attachTypeInfo(node: CheckerTreeRuntimeNode<T>) {
-		if (node.validDefinitions.length === 1) {
-			const { typeDef, desc } = node.validDefinitions[0]
-			options.attachTypeInfo(node.node.originalNode, typeDef, desc)
-			handleAttributes(typeDef.attributes, options.context, (handler, config) => {
+		function handleStringAttachers(
+			runtimeValue: RuntimeNode<T>,
+			typeDef: SimplifiedMcdocTypeNoUnion,
+		) {
```
misode, line 457: This function, and now that I think about it, maybe also `attachTypeInfo` could be moved out to the top level of the module. They only need an additional `options` parameter, they don't capture anything else right now.

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -691,9 +708,15 @@ function condenseErrorsAndFilterSiblings<T>(
 	return { definitions: validDefinitions.map(d => d.definition), condensedErrors: errors }
 }
 
+interface ValidDefintionResultChildDefintion {
```
misode, line 711: This is a mouthful :D

NeunEinser: I renamed it, so it dosen't have `definition` twice anymore at least.

misode (approved):

PR #1314, opened 2024-06-17
🐛 Fix completer issue when cursor is not inside string node

misode:
When the completer cursor is not inside the string node, don't try to get the completions from it, instead fall back to the `getValues` call, which corrects the range to the current cursor.

This didn't seem to be needed for nbt, but I added it anyways to keep the code consistent. It seems like the real issue is caused by the JSON parser, adding an empty `json:string` node for pairs. This should be corrected in the future. But even with that fix, it doesn't hurt to have this check.

### DIFF
--- a/packages/json/src/completer/index.ts
+++ b/packages/json/src/completer/index.ts
@@ -62,14 +62,14 @@ const object = core.completer.record<JsonStringNode, JsonNode, JsonObjectNode>({
 })
 
 const primitive: core.Completer<JsonPrimitiveNode> = (node, ctx) => {
-	if (node.type === 'json:string' && node.children?.length) {
+	const insideRange = core.Range.contains(node, ctx.offset, true)
+	if (node.type === 'json:string' && node.children?.length && insideRange) {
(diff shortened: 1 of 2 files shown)


NeunEinser:
One small issue is that it doesn't replace a previously existing value when completing outside of the string

Peek 2024-06-18 00-31

misode:
Hmm. What should be the expected result here?

NeunEinser:
I think, ideally it would completely override the existing string node like it happens when selecting `border` in the example

misode:
> I think, ideally it would completely override the existing string node like it happens when selecting `border` in the example

I believe when vscode receives a completion item that doesn't overlap with the `ctx.offset`, it is discarded. We'd have to do weird hacks with completion item's `additionalTextEdits`, I'm not sure that's worth it.

NeunEinser (approved):

NeunEinser:
We also have this case where we wouldn't want that, so yeah.

Peek 2024-06-18 00-51

PR #1330, opened 2024-06-18
✨ Add validation when checking `count` in item predicates

### LINKED ISSUE #1303: Support special case `count` as item predicate test
`count` is not an item sub predicate (it isn't allowed in JSON files), but it is a special case in the item predicate argument.

Trivaxy:
This PR adds validation to the `count` component in item predicates.

The following is implemented:
- Checking only whether `count` exists or not produces a warning, as it is useless
- Checking if `count` is a specific value produces an error if the value is not positive, e.g. `count=-1`
- Any irrelevant keys produce warnings, e.g. `count={min:2,foo:[]}` will warn on `foo`
- If `min` or `max` are specified more than once then errors are produced, e.g. `count={min:2, min:3, max: 4}`
- If no keys are specified, a warning is produced, i.e. `count={}`
- If `min` is greater than `max`, an error is produced, i.e. `count={min:10, max:2}`
- If `min` or `max` are negative, an error is produced, i.e. `count={min:-2}`
- Checking `count`, `min`, or `max` against an invalid nbt type will produce an error

Many of the things caught here function fine in vanilla, but it's good to catch them anyways. The choice between reporting a warning vs. an error was made based on if the command will behave strangely - for example, an existence check for `count` is harmless so it gets a warning, but specifying multiple `min`/`max` keys has the unexpected behavior of the game only taking into account the last duplicate specified, so in this case an error is thrown by us to save the user from headaches.

Fixes #1303


packages/java-edition/src/mcfunction/parser/argument.ts
```diff
@@ -1533,6 +1533,13 @@ const componentTest: core.InfallibleParser<ComponentTestNode> = (src, ctx) => {
 	src.skipWhitespace()
 
 	const key = core.resourceLocation({ category: 'data_component_type' })(src, ctx)
+
+	// hack: if the key is minecraft:count, change the node's options. there has to be a better way to do this
```
misode, line 1537: I think this is a pretty good way of doing it tbh

packages/java-edition/src/mcfunction/checker/index.ts
```diff
@@ -97,7 +98,108 @@ const itemPredicate: core.SyncChecker<ItemPredicateNode> = (node, ctx) => {
...
+					if (ComponentTestExistsNode.is(test)) {
+						ctx.err.report(
+							localize('mcfunction.checker.item-predicate.useless-count-check'),
+							test.range,
+							core.ErrorSeverity.Warning,
+						)
+					} else if (
+						(ComponentTestExactNode.is(test) || ComponentTestSubpredicateNode.is(test))
+						&& test.value
+					) {
```
misode, line 113: The whole body of this if statement should be checked by mcdoc ideally. But I understand you did not want to do that in this PR. We should not have these very specific error messages for every use case.

misode (approved):
This can definitely be simplified with mcdoc in the future, but for now it seems to work

PR #1334, opened 2024-06-18
🐛 Implement `isPredicate` NBT checks

### LINKED ISSUE #1317: Implement `isPredicate` in NBT checker
* All keys should be optional (yes, even required fields from codecs), because this uses Minecraft's legacy partial NBT matcher
* Numeric types should always have their suffix, ie. shouldn't be assignable to each other
* The `minecraft:` namespace prefix can't be inferred.
* .... possibly more?

misode:
- Fixes #1317 

image


misode:
Merging this, because the context refactor is going to cause conflicts with #1336 otherwise

PR #1335, opened 2024-06-19
🐛 Fix: also filter spread fields

misode:
Only pair fields were getting filtered by attributes like `since` and `until`. Moving the logic to an earlier place.

### DIFF
--- a/packages/mcdoc/src/runtime/checker/index.ts
+++ b/packages/mcdoc/src/runtime/checker/index.ts
@@ -1245,19 +1245,11 @@ function simplifyStruct<T>(typeDef: StructType, context: SimplifyContext<T>): Si
 	let complexFields: SimplifiedStructTypePairField[] = []
 
 	function addField(key: string | SimplifiedMcdocType, field: StructTypePairField) {
-		let keep = true
 		handleAttributes(field.attributes, context.options.context, (handler, config) => {
-			if (!keep) return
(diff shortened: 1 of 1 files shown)


packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -1280,6 +1272,15 @@ function simplifyStruct<T>(typeDef: StructType, context: SimplifyContext<T>): Si
 		}
 	}
 	for (const field of typeDef.fields) {
+		let keep = true
+		handleAttributes(field.attributes, context.options.context, (handler, config) => {
+			if (keep && handler.filterElement?.(config, context.options.context) === false) {
```
TheAfroOfDoom, line 1277: does `filterElement` not for sure return `false`? i guess it can be `undefined` here but it's not immediately clear to me why `undefined` wouldn't also make `keep = false`

not that this logic is new to this PR

misode: `filterElement` always returns `true` or `false`, but it will be `undefined` if `handler.filterElement` is undefined. Attributes that don't have this handler method shouldn't influence whether a field is kept or not.

misode: I have plans to refactor `handleAttributes` slightly so that it takes a method string (`'filterElement` in this case) and only loops over the attributes that have this method. Both to improve performance and to make things slighly easier to read.

PR #1336, opened 2024-06-19
🐛 Fix issues with `id` attribute

### LINKED ISSUE #1262: Add support for `id` attribute on enums
Example: loot table `type` field

misode:
- Fixes #1262 
- Fixes `#[id] string`
- Fixes `id` attributes on literal strings


MulverineX (approved):

misode:
I think this is done now. Everything seems to work as expected. Only problem now is that the item component struct is very expensive to calculate, due to all the complex fields. A json file like this takes 300ms to check:
```json
{
  "condition": "match_tool",
  "predicate": {
    "components": {
      "attribute_modifiers": [],
      "minecraft:attribute_modifiers": []
    }
  }
}
```

misode:
After making some changes in mcdoc, we no longer have any `id` attributes on literals, but these changes shouldn't hurt. The fixes for enums are still definitely necessary.

PR #1337, opened 2024-06-19
🐛 Fix getAllFiles nodejs implementation on windows

misode:
The `globby` expects only `/`, but `toPath` returned `\` on windows. Implementation is a bit rough, there may be other uses of `toPath` that need this treatment.


SPGoding (approved):
I guess it gets the job done. Might be worth it to remove `globby` and do our own implementation at some point.

PR #1338, opened 2024-06-19
🐛 Add patch for schedule clear command

misode:
The game uses a greedy string to parse this argument, with a custom suggest method that lists all the currently scheduled functions. We need to patch this to just show all functions.


SPGoding (approved):
Might be nice if we only suggest functions that are scheduled at some point, at some point.

PR #1340, opened 2024-06-19
🐛 Fix symbol clear not checking contributor when removing locations

### LINKED ISSUE #1327: Recursive functions are sometimes not found
image

misode:
- Fixes #1327 

The `SymbolUtil#clear` function takes a `contributor`, but the bug was that it only used this contributor for getting the right symbol cache. When actually removing the symbol locations it removes all of the locations. This breaks when there is a symbol reference (contributed by `binder`) in the same file that defines it (contributed by `uri_binder`).


SPGoding (approved):
Nice catch

PR #1342, opened 2024-06-20
🐛 Make component tests between brackets optional

### LINKED ISSUE #1331: Empty item predicates shows an error
Example: `*[]` will toss out an error that a resource location is expected, but this is acceptable syntax in vanilla.

misode:
I caused this in #1302, it's only fair that I fix it again

- Fixes #1331


TheAfroOfDoom (commented):
it'd be really nice if we had tests for this sorta thing so we could see the behavior change super easily (error -> not error or something)

PR #1346, opened 2024-06-20
✨ Fix missing item `id` when attempting to index `custom_item_data`

MulverineX:


packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -877,18 +877,14 @@ function simplifyIndexed<T>(
...
 		return { kind: 'union', members: [] }
 	}
 	let values: McdocType[] = []
 
 	for (const index of typeDef.parallelIndices) {
 		let lookup: string[] = []
 		if (index.kind === 'static') {
-			if (index.value === '%fallback') {
-				values = child.fields.filter(f => f.kind === 'pair').map(f => f.type)
-				break
```
misode, line 890: Is this removal intended? Does `foo: minecraft:item[%fallback]` still work?

packages/java-edition/src/mcfunction/checker/index.ts
```diff
@@ -217,13 +217,31 @@ const itemStack: core.SyncChecker<ItemStackNode> = (node, ctx) => {
...
-		groupedComponents.get(key)!.push(pair.key)
+		groupedComponents.get(componentId)!.push(pair.key)
 		if (pair.value) {
-			nbt.checker.index('minecraft:data_component', key)(pair.value, ctx)
+			if (componentId === 'minecraft:custom_data') {
+				if (pair.value.type === 'nbt:string') {
+					// TODO: Maybe move this to the nbt package
+					const stringNBT = nbt.parser.compound(
+						new core.Source(pair.value.value, pair.value.valueMap),
+						ctx,
```
misode, line 231: Yeah... it's not ideal that it has to run the parser here. The `#[nbt=mcdoc:custom_item_data[[%parent.id]]] string` mcdoc type should already be parsing the string as an NBT compound.

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -979,9 +975,17 @@ function simplifyIndexed<T>(
...
 		if (currentValues.includes(undefined)) {
-			// fallback case
-			values = child.fields.filter(f => f.kind === 'pair').map(f => f.type)
-			break
+			const fallbackDispatch = child.fields.find(f =>
+				f.kind === 'pair' && f.key.kind === 'literal' && f.key.value.value === '%fallback'
+			)
+
+			if (fallbackDispatch) {
+				values.push(fallbackDispatch.type)
```
misode, line 983: What are these changes supposed to achieve? Why was this necessary?

packages/java-edition/src/mcfunction/checker/index.ts
MulverineX: It does, but not in this context. Also, the `nbt` attribute doesn't actually check against the definition yet.

packages/mcdoc/src/runtime/checker/index.ts
MulverineX: Because otherwise, it creates a union of all dispatched types, rather than using the defined `%fallback`

packages/java-edition/src/mcfunction/checker/index.ts
misode: The `nbt` attribute does run the checker...

packages/mcdoc/src/runtime/checker/index.ts
MulverineX: @NeunEinser would know, he said I should remove it

MulverineX: (Answered in Discord)

packages/java-edition/src/mcfunction/checker/index.ts
MulverineX: (its kinda bugged, there was discussion in discord)

misode (approved):
Seems to work well. The string parsing can be refactored in the future. I'll trust what neun was doing while pair-coding the fallback changes 👍

PR #1352, opened 2024-06-20
🐛 Fix cached files in archives not detected as removed

### LINKED ISSUE #1213: Failed creating TextDocument for “archive://mcmeta.tar.gz/data/...” Error: Archive “mcmeta.tar.gz” has not been loaded into the memory
Warnings like these are spammed in the output log during the "project#ready bind files" phase (taking up to 3 minutes, but that may be caused by the output log being open)
```
[Warn  - 4:31:52 PM] [Project] [read] Failed creating TextDocument for “archive://mcmeta.tar.gz/data/minecraft/banner_pattern/diagonal_right.json” Error: Archive “mcmeta.tar.gz” has not been loaded into the memory
    at _ArchiveUriSupporter.getDataInArchive (<project>\packages\core\src\service\FileService.ts:311:10)
    at _ArchiveUriSupporter.readFile (<project>\packages\core\src\service\FileService.ts:299:15)
    at FileServiceImpl.readFile (<project>\packages\core\src\service\FileService.ts:133:41)
    at createTextDocument (<project>\packages\core\src\service\Project.ts:719:50)
    at getCacheHandlingPromise (<project>\packages\core\src\service\Project.ts:754:21)
    at _Project.read (<project>\packages\core\src\service\Project.ts:785:10)
    at _Project.ensureBindingStarted (<project>\packages\core\src\service\Project.ts:912:26)
    at ready (<project>\packages\core\src\service\Project.ts:622:16)
    at _Project.ready (<project>\packages\core\src\service\Project.ts:647:3)
    at <anonymous> (<project>\packages\language-server\src\server.ts:192:2)
    ```

misode:
Cached files in an archive were getting detected as "changed" rather than "removed". The issue was that the cache service relied on detecting `ENOENT` and `EISDIR` codes in the thrown error, but this only happened for actual files on the file system, not for files in an archive.

Importantly, this does **not** solve #1213, I haven't been able to reproduce that error (which is slighly different).

### DIFF
--- a/packages/core/src/common/externals/BrowserExternals.ts
+++ b/packages/core/src/common/externals/BrowserExternals.ts
@@ -189,6 +189,9 @@ export const BrowserExternals: Externals = {
 	},
 	downloader: new BrowserExternalDownloader(),
 	error: {
+		createKind(kind, message) {
+			return new Error(`${kind}: ${message}`)
+		},
 		isKind(e, kind) {
 			return e instanceof Error && e.message.startsWith(kind)
 		},
--- a/packages/core/src/common/externals/NodeJsExternals.ts
+++ b/packages/core/src/common/externals/NodeJsExternals.ts
@@ -68,6 +68,11 @@ export const NodeJsExternals: Externals = {
 	},
 	downloader: new NodeJsExternalDownloader(),
 	error: {
+		createKind(kind, message) {
+			const error = new Error(message)
+			;(error as NodeJS.ErrnoException).code = kind
+			return error
+		},
 		isKind(e, kind) {
 			return e instanceof Error && (e as NodeJS.ErrnoException).code === kind
 		},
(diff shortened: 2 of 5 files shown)


misode:
@SPGoding What should happen when it reaches the "has not been loaded into the memory" error? Should this count as a file removal or should it stay like the current behavior?

SPGoding:
Is it possible to add a function to `Externals` for creating IO Errors of a certain kind instead of using a costom `Error` type?

As for the "has not been loaded into memory" errors, I feel like they should be kept as-is for now until we can find out what's causing them since it shouldn't really happen.

MulverineX (approved):

misode:
> Is it possible to add a function to `Externals` for creating IO Errors of a certain kind instead of using a costom `Error` type?

Good suggestion, looks much cleaner now!

SPGoding (approved):

PR #1353, opened 2024-06-20
♻️ Add `PrefixedNode` for prefixed resource locations

misode:


packages/core/test/parser/prefixed.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { describe, it } from 'mocha'
+import snapshot from 'snap-shot-it'
+import type { BooleanNode } from '../../lib/index.js'
+import { boolean } from '../../lib/index.js'
+import { type Options, prefixed } from '../../lib/parser/prefixed.js'
+import { showWhitespaceGlyph, testParser } from '../utils.js'
+
+describe.only('prefixed()', () => {
```
TheAfroOfDoom, line 8: `.only`

TheAfroOfDoom (commented):
whats the point of this refactor?

misode:
> whats the point of this refactor?

Right now when you autocomplete components in `set_components`, the prefixed and non-prefixed fields both show up without a `!`. This is because the `id` attribute needs to return a mock AST node, which is used for completions.
image

NeunEinser (approved):

Issue #1296, opened 2024-06-16
Index map issue with nested strings


misode:
This needs to be investigated

```
summon cow ~ ~ ~ {CustomName:'{"text":"hello","hoverEvent":{"action":"show_item","value":"{id:\'acacia_boat\'}"}}'}
```

![image](https://github.com/SpyglassMC/Spyglass/assets/17352009/7008c366-8380-49a2-a3ff-63b4b961324a)

misode:
I found another case where it went wrong:
![image](https://github.com/SpyglassMC/Spyglass/assets/17352009/4e66fa98-1bbb-498f-b2cf-b8d460f20bcc)
Interestingly, it works correctly when using only double quotes:
![image](https://github.com/SpyglassMC/Spyglass/assets/17352009/009cd5eb-617d-438b-8b7f-f32aa20b89b0)

TheAfroOfDoom:
from https://github.com/SpyglassMC/Spyglass/pull/1355#issue-2365408905

[...]

PR #1355, opened 2024-06-20
✨ Add debug method to `Source` for better `IndexMap` visualization

TheAfroOfDoom:
this was helping so far in troubleshooting #1296. specifically in figuring out what point the index map was being messed up.

see snapshots

---

sample use with console.logs:

image


misode (approved):
Looks pretty useful and is entirely separated from the rest of the code, so completely safe!

PR #1356, opened 2024-06-21
🐛 Fix nested binders and checkers when direct child cannot dispatch

misode:


NeunEinser (commented):
I don't understand the issue being fixed here, so obligatorily lgtm I suppose? :D

misode:
Previously the `core.binder.dispatchSync(node, ctx)` and `core.checker.dispatchSync(node, ctx)` calls at the end of the string attachment would only shallowly look at the children of the `node` and see if there was a binder and checker registered for those types. If that was not the case it wouldn't do anything, this was affecting for example the command attribute. I changed it so it now finds the shallowest children that do have a binder/checker registered.

SPGoding (approved):
There is a different function `AstNode.findShallowestChild()` that does a similar thing to how `traversePreOrder()` is being used here that might make the code easier to follow? Tbh no idea why these two functions are separately implemented.

misode:
I think keeping the implementation of `fallback` and `fallbackSync` is the cleanest. We can in the future look into combining these two implementations.

PR #1359, opened 2024-06-21
✨ Add completions for empty strings, structs, and lists

misode:


TheAfroOfDoom (commented):
not to be too annoying about it but either a test _or_ a blurb/example in the PR description would be helpful for those of us not super familiar with the intention/feature/bug etc. in-action (me)

misode:
We don't have any tests for completions right now, but I can attach a screenshot, sure 🙂

PR #1366, opened 2024-06-21
🐛 Bind enchantment provider

jacobsjo:
Enchantment providers were missing from the binder.


misode (approved):

MulverineX (approved):

MulverineX:
Ah yep this is my bad, good catch 👍

PR #1367, opened 2024-06-22
♻️ Fix mcdoc key definition not being linked to the value defintion and handle definition being empty union

### LINKED ISSUE #1357: Mcdoc key attacher can run for invalid definitions
During the `attachTypeInfo` process, when it reaches the struct inside `enchantments`, the type checker has narrowed the valid type definitions down to one (because `{"aqua_affinity": 1}` is not an int).

Hovering over `aqua_affinity`, or getting autocompletions for `show_in_tooltip` makes it clear that the type info is correctly attached to values. However, when it attaches the type info for the key and runs the string attachers of the key, it seems to use the wrong type definition.

```json
{
  "function": "minecraft:set_components",
  "components": {
    "enchantments": {
      "levels": {
        "aqua_affinity": 1
      }
    }
  }
}
```
image

NeunEinser:
Fixes #1357

This restructured the checker tree again, this time to have a list of simplified definitions per parent definition.

Each definition will have a unique mapping for a simplified typedef to each runtime child. However, this typedef could be a union, allowing multiple alternatives for the child.

This PR adds an additional grouping of these definitions, so we will have a unique group per possible parent definition.

This has following advantages:
- Allows us to keep track of when a parent has an empty union as definition for a child
- This avoids duplicating each key definition for every valid value definition, in case the key value type is defined as a union. Like this, each key definition is only once in the tree and no additional deduplication is needed*
- Calling error condensing is a bit more natural and doesn't require additional filtering based on the parent, since everything is already grouped by parent anyway

\* You could still have multiple definitions defining the key with an equivalent definition, in which case both equivalent definitions would be attached to the key. This is the same behavior as for values though, unions are currently not deduplicated, so this is out of scope for this PR.


misode:
This change seems to have caused the following error: for literal keys it no longer shows the mcdoc type of the value of the field.

Right now I think it's fine to compromize and say that literal keys will alway show the type of the value, and complex keys will show their own type. But In the future we should in the latter case also show the type of the value.

image

NeunEinser:
> This change seems to have caused the following error: for literal keys it no longer shows the mcdoc type of the value of the field.
> 
> Right now I think it's fine to compromize and say that literal keys will alway show the type of the value, and complex keys will show their own type. But In the future we should in the latter case also show the type of the value.
> 
> image

Why not show the tooltip with the typedef on the value instead? I think that would be a bit cleaner, at least this is where I was expecting it. I was actually checking this on my branch and thought "ah well I guess it's not implemented".

NeunEinser:
Hmm, I kind of like this
image

However, when hovering in some gap that doesn't have hover info, it uses the hover from the struct, which isn't great.
image

NeunEinser:
Okay, my best guess why this was working before is maybe the execution order might have been different and before it set the keydef and then the value def and now it's switched?

Anyways, I got this working now by just waiting for both the keydef and the value def, and the setting the combined thing on the pair itself rather than the key, and than additionally setting the value to just the value def. I kind of like this, there are just some weird cases when hovering over tabs. The gaps are shown correctly because they are part of the pair.
Peek 2024-06-22 23-55

misode:
I'm not a fan of this new hover behavior. I'm mostly basing this on how other language servers work. When you hover over a key it shows the type of the value of that pair:
image

NeunEinser:
What do you not like about it? It is providing info on more places than before. I think it's pretty useful, since you also have a consistent way of getting information and hover texts on e.g. list elements where you do not have a key.

If the only thing you dislike is the weird behavior with whitespace, this would be completely fixed if I do not attach type info to lists and objects, and only add the hover to primitive types.

NeunEinser:
> [Remove hover from pair nodes and primitive values](https://github.com/SpyglassMC/Spyglass/pull/1367/commits/1cef56dad4e40ea32fa1d1a7f191eaec1b0b2d60)

This commit message should have said **non**-primitive values

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -192,187 +191,199 @@ export function typeDefinition<T>(
...
-							(simplified.kind === 'union' ? simplified.members : [simplified]).map(d => ({
+			for (const definitionGroup of value.definitionsByParent) {
+				const definitionErrors: ErrorCondensingDefinition<T>[] = []
+
+				if (definitionGroup.validDefinitions.length === 0) {
+					// nothing can be assigned to an empty union
+					definitionGroup.condensedErrors = [[{
+						kind: 'type_mismatch',
+						node: value.node,
+						expected: [],
```
misode, line 255: This is not accounted for in the error reporter

misode (commented):
I didn't really understand how `typeDefinition` worked before this so I'm not going to try to understand 😄 

Everything seems to work great though!

packages/mcdoc/src/runtime/checker/index.ts
NeunEinser: I don't think it needs to have a special check?

I checked the implementation of `arrayToMessage` and it looks like it returns a translated `nothing` when an empty array is supplied.

I haven't tested this though.

misode: Oh it does? Neat! my bad

misode (approved):

PR #1369, opened 2024-06-22
🧑‍💻 Add benchmarks for mcdoc runtime checker

misode:
Adds a new package `benchmarks` using [tinybench](https://github.com/tinylibs/tinybench). After building, you can run `npm run bench` in the root folder to run the benchmarks.

Benchmarks do not rely on vanilla-mcdoc so that they can be stable. Although it might be useful to add a more real case that actually uses vanilla-data to do a real performance test.

image


NeunEinser (approved):

SPGoding (commented):
Any reason we don't save the `benchmarks` package in `.packages.json`?

misode:
> Any reason we don't save the `benchmarks` package in `.packages.json`?

I don't think we want to release this package to npm, and it was throwing errors when running the release script.

MulverineX (approved):

PR #1370, opened 2024-06-22
⚡️ Improve mcdoc struct spread simplify performance

NeunEinser:
Before:
image
After:
image

Also fixed a small bug that all union members would always call simplify, even if they had an attribute that would mean they are filtered in the current context.


packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -994,45 +1003,55 @@ function simplifyIndexed<T>(
...
-		if (simplified.kind === 'union') {
-			members.push(...simplified.members)
-		} else {
-			members.push(simplified)
-		}
+	let memberContext = context
+	if (context.literalStructFields && validMembers.length > 1) {
+		memberContext = { ...context, spreadingUnion: true }
 	}
+	const members = validMembers.map(member => simplify(member, memberContext))
```
NeunEinser, line 1028: Actually now that I think about it, this will not work 100% correctly when one of the types resolves to an empty union later (e.g. if there are two valid members, one of which was a reference which itself resolves to an empty union).

NeunEinser: Fixed, without meaningful slowdown
image

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -874,7 +877,13 @@ function simplifyIndexed<T>(
 	typeDef: IndexedType,
 	context: SimplifyContext<T>,
 ): SimplifiedMcdocType {
-	const child = simplify(typeDef.child, { ...context, typeArgs: [] })
+	const child = simplify(typeDef.child, {
+		...context,
+		typeArgs: [],
+		spreadingUnion: false,
+		literalStructFields: undefined,
+		complexStructFields: undefined,
```
misode, line 885: Is this the only place where you'd want to reset these context variables?

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -994,45 +1003,75 @@ function simplifyIndexed<T>(
...
+	const members = validMembers
+		.flatMap(member => {
+			const simplified = simplify(member, { ...context, spreadingUnion: true })
+			if (simplified.kind === 'union') {
+				return simplified.members
+			}
+			return [simplified]
+		})
+
+	if (context.literalStructFields && !context.spreadingUnion) {
```
misode, line 1033: Is it intended that complex fields are not handled here?

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -889,11 +878,47 @@ function simplifyIndexed<T>(
...
 		return { kind: 'union', members: [] }
 	}
+
+	const symbolMap: { [key: string]: { data?: unknown } } = {}
+	for (const field of child.fields) {
+		if (field.key.kind === 'literal' && field.key.value.kind === 'string') {
+			symbolMap[field.key.value.value] = {
+				data: {
+					typeDef: field.type,
+				} satisfies TypeDefSymbolData,
```
misode, line 888: It feels weird re-using the `TypeDefSymbolData` for non-dispatcher types tbh, but maybe it's the cleanest solution.

packages/mcdoc/src/runtime/checker/index.ts
NeunEinser: This is only for marking fields optional that have been added by this union.
Complex fields are always optional, so they neither set the `unionMember` bool, nor do they need to be iterated here.

packages/mcdoc/src/runtime/checker/index.ts
NeunEinser: An alternative would be to map the type when using a dispatcher, but since dispatcher is the default case, I wanted to try to not do any mapping on the dispatcher.

It probably wouldn't change much though.

packages/mcdoc/src/runtime/checker/index.ts
NeunEinser: That was an obvious one, as we don't want to add the fields of a type we index into added to the parent struct.

I can't think of another case where it would be required, but I could be missing something.

NeunEinser: With the new approach, I am also doing this now when resolving a union member.

packages/mcdoc/src/runtime/checker/index.ts
NeunEinser: Resolving as it's no longer used.

NeunEinser:
New benchmark; I am not sure why the trivial case improved so much.
image

misode (approved):

misode:
Impressive results! Checking the large mcfunction file that someone shared on Discord went from ~900ms -> 400ms!

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -856,35 +860,64 @@ function simplifyDispatcher<T>(
...
+			const data = symbolMap['%unknown']?.data
+			unkownTypeDef = TypeDefSymbolData.is(data) ? data.typeDef : undefined
+		}
+		return unkownTypeDef
+	}
+
+	for (const index of parallelIndices) {
 		let lookup: string[] = []
 		if (index.kind === 'static') {
+			if (index.value === '%fallback') {
```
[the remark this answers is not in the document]
NeunEinser: Yes this is precisely why I re-added it.

This code will run when explicitly trying to access %fallback. Without this, %fallback would act like %unknown, as it would try to find an explicitly dispatched fallback, not find anything, and then run the %unknown case.

So this was correct before, I just forgot about %unknown when I was in voice with Mulv.

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -1081,31 +1140,27 @@ function simplifyStruct<T>(typeDef: StructType, context: SimplifyContext<T>): Si
...
+			if (simplifiedSpread.kind === 'any') {
+				addField({ kind: 'any' }, { kind: 'pair', key: { kind: 'any' }, type: { kind: 'any' } })
 			}
-			// Silently ignore spreading non-struct types
 		}
 	}
+	if (context.structFields) {
+		// In this case we are spreading a struct and the fields have been added to the parent map
+		// from context now.
+		return { kind: 'struct', fields: [] }
```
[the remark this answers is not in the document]
NeunEinser: Yeah, I also kind of dislike this. I am also not sure how this would work with caching. We would definitely cache the intermediate result of a reference, even when used as a spread operator, which isn't really possible with this anymore.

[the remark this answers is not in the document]
NeunEinser: Actually, now that I think about it, the whole deduplicating is likely not needed anymore. The shallow check already constructs a map of literal keys of the runtime value, so if multiple keys in the simplified struct type match, I think it would just overwrite the definition multiple times, keeping the last one. I'll check that after work, that'd probably the best case, both for code abd for performance.

The only case I am not sure about is complex keys

PR #1372, opened 2024-06-22
⚡ Filter .gitignore and add `env.exclude` config

### LINKED ISSUE #970: Add possibility to exclude file paths
On big datapacks that use systems such as search trees, the number of file is huge, and then the cache take hours to update... each time we load the datapack in VSC. It result of a high battery usage and a massive performance issue on the plugin (syntax highlight can take entire minutes to display correctly).

It would be great to be able to disable the cache, or just add the opened files to the cache.

### LINKED ISSUE #1308: Ignore files in .gitignore by default

### LINKED ISSUE #1371: Don't track unnecessary files
Right now the language server will track and parse all files as long as they match a set of file extensions (even files like `.vscode/settings.json`)

Even without a user explicitly defining an ignore list or specifying a gitignore (#1308), we should filter out these unnecessary files.

misode:
- Fixes #970 
- Fixes #1308 
- Fixes #1371 

I'm not sure if this was the right place to filter files. For example chokidar will still watch all files even if they're in gitignore.


NeunEinser:
Instead of ignoring .vscode, would it make sense to allow list `data` and `assets` directories instead?

Or is this for the case a .vscode is inside of a data pack?

TheAfroOfDoom:
> Instead of ignoring .vscode, would it make sense to allow list `data` and `assets` directories instead?

i think ignoring `.vscode` by default makes sense and i'm all for it!

next steps would probably be to add a config setting so users can specify their own paths to ignore.

misode:
> Instead of ignoring .vscode, would it make sense to allow list `data` and `assets` directories instead?

That is not easy to do. What if you have a folder `node_modules/foo/data/bar`? Should it also look at files in here? Ignore patterns are usually much easier to work with. I think I'll add an explicit `env.ignore` config in this PR as well.

packages/core/src/service/Project.ts
```diff
@@ -390,17 +389,22 @@ export class Project implements ExternalEventEmitter {
...
+					try {
+						const uri = this.projectRoot + Project.GitIgnore
+						const contents = bufferToString(await this.externals.fs.readFile(uri))
+						this.ignore.add(contents)
+					} catch (e) {
+						if (this.externals.error.isKind(e, 'ENOENT')) {
+							continue
+						}
+						this.logger.error(`[Project] [Config] Failed reading .gitignore`, e)
+					}
```
TheAfroOfDoom, line 404: this seems like a good candidate to be refactored into a separate method for readability

`private parseGitIgnore()` maybe

TheAfroOfDoom (approved):

TheAfroOfDoom:
do the ignored files still get parsed for symbols? e.g. function references

misode:
> do the ignored files still get parsed for symbols? e.g. function references

No, ignored files are excluded from the parse+bind phase that happens in the ready phase. This is the major speedup when opening a workspace with large ignored folders.

packages/core/src/service/Project.ts
misode: Done!

PR #1377, opened 2024-06-23
🐛 Allow missing keys in data merge command

### LINKED ISSUE #1373: Required fields may already be present on a `/data` target; merging data doesn't need to be complete
For example:
image

These fields are required when summoning a display entity, but here someone is trying to only modify some of the fields

misode:
- Fixes #1373


misode:
Even though there hasn't been a review, this is a fairly trivial implementation and it's been a day.

PR #1378, opened 2024-06-23
✨ Support multiple workspace folders

### LINKED ISSUE #1354: Support multi-folder workspaces
image

Posted by `@minihilly` on discord, only the first pack, `Core` is being handled by the extension

misode:
This is a naive implementation of supporting multiple workspace folders. Just because this PR exists doesn't mean that we will support multiple workspace folders soon.
- Closes #1354 

### Open questions
* [x] The `CacheService` currently uses `projectRoots[0]` to compute the `getCacheFileUri`. This is probably not a good idea, since it means the symbol cache of the first project root will also contain symbols of the other project roots. Either a unique key should be used for each combination of workspace folders, or each project root should have their own symbol cache.
  * Fixed, it now combines the project roots to compute the cache file uri
* [ ] Right now the list of project roots is just passed to a single project and the file watcher will watch all those roots. Should we instead create a `Project` for each project root?
* [x] The `spyglass.json` config file is only loaded in the first project root. This is probably okay, although the config documentation indicates that `lint` and `format` rules may be specified on a project root level granularity.
  * Fixed, it now merges configs in all workspace folders 
* [x] Only the `pack.mcmeta` of the first workspace folder is loaded. But reading this, it seems it doesn't even work for nested pack.mcmetas so we might need to change this anyways.
  * Fixed, it now searches through all workspace folders


packages/core/src/common/externals/BrowserExternals.ts
```diff
@@ -148,7 +148,7 @@ class BrowserFileSystem implements ExternalFileSystem {
 		delete this.states[location]
 		this.saveStates()
 	}
-	watch(_location: FsLocation): FsWatcher {
+	watch(_location: FsLocation[]): FsWatcher {
```
SPGoding, line 151: nit (same for the implementation signatures): 
```suggestion
	watch(_locations: FsLocation[]): FsWatcher {
```

packages/core/src/service/CacheService.ts
```diff
@@ -114,7 +114,9 @@ export class CacheService {
 	 */
 	private async getCacheFileUri(): Promise<string> {
 		return (this.#cacheFilePath ??= new Uri(
-			`symbols/${await this.project.externals.crypto.getSha1(this.project.projectRoot)}.json.gz`,
+			`symbols/${await this.project.externals.crypto.getSha1(
+				this.project.projectRoots[0],
```
SPGoding, line 118: I think using a combination of all roots here would be appropriate (something like `projectRoots.sort().join(':')`? ). Imo the expected behavior for multi-root is that we can reference resources across roots, so the same symbol table should be shared across (and therefore cached for) all roots. Though we'll have to think about how to support multi version properly.

packages/core/src/service/CacheService.ts
```diff
@@ -219,7 +221,7 @@ export class CacheService {
 			filePath = await this.getCacheFileUri()
 			const cache: CacheFile = {
 				version: LatestCacheVersion,
-				projectRoot: this.project.projectRoot,
+				projectRoot: this.project.projectRoots[0],
```
SPGoding, line 224: Can we store all roots here? I can't remember what is this used for though.

SPGoding (commented):
For the config file and `pack.mcmeta` I think we can just leave them like this for now.

packages/core/src/service/CacheService.ts
misode: Afaict this field isn't used

SPGoding (approved):

PR #1380, opened 2024-06-23
:bug: Fix selector parser being too strict in /say

### LINKED ISSUE #1304: `/say` selector parser is too strict
It shouldn't error for cases like this. Reference the decompiled code to see how Minecraft parses this.

image

Calverin:
Fixes #1304

### DIFF
--- a/packages/java-edition/src/mcfunction/parser/argument.ts
+++ b/packages/java-edition/src/mcfunction/parser/argument.ts
@@ -496,8 +496,8 @@ const message: core.InfallibleParser<MessageNode> = (src, ctx) => {
 	}
 
 	while (src.canReadInLine()) {
-		if (src.peek() === '@') {
-			ans.children.push(selector()(src, ctx) as EntitySelectorNode)
+		if (EntitySelectorAtVariable.is(src.peek(2))) {
+			ans.children.push(selector(true)(src, ctx) as EntitySelectorNode)
 		} else {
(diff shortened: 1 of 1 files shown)


packages/java-edition/src/mcfunction/parser/argument.ts
```diff
@@ -679,10 +679,16 @@ function resourceOrInline(category: core.FileCategory) {
...
 
-function selectorPrefix(): core.InfallibleParser<core.LiteralNode> {
+function selectorPrefix(ignoreInvalidPrefix: boolean): core.InfallibleParser<core.LiteralNode> {
 	return (src: core.Source, ctx: core.ParserContext): core.LiteralNode => {
 		const start = src.cursor
-		const value = src.readUntil(' ', '\r', '\n', '[')
+		let value: string
+		if (ignoreInvalidPrefix) {
+			value = src.peek(0, 2)
+			src.skip(2)
```
TheAfroOfDoom, line 688: i think this will do the same thing for you:

```suggestion
			value = src.read(2)
```

Calverin: `src.read()` doesn't accept any arguments, I had thought this would work as well but it just doesn't exist. Mandela effect something something...

MulverineX (approved):

misode (approved):

PR #1384, opened 2024-06-23
✨ Add THIS completion item

### LINKED ISSUE #1329: Re-add `THIS` autocomplete entry

misode:
- Closes #1329

image
image

### DIFF
--- a/packages/core/src/processor/completer/builtin.ts
+++ b/packages/core/src/processor/completer/builtin.ts
@@ -211,15 +211,36 @@ export const resourceLocation: Completer<ResourceLocationNode> = (node, ctx) => 
 	const pool = node.options.pool
 		? optimizePool(node.options.pool)
 		: [
-			...getPool(node.options.category!),
+			...getPool(node.options.category),
 			...(node.options.allowTag
-				? getPool(`tag/${node.options.category}` as TagFileCategory).map((v) =>
+				? getPool(`tag/${node.options.category}`).map((v) =>
 					`${ResourceLocation.TagPrefix}${v}`
 				)
 				: []),
 		]
 
-	return pool.map((v) => CompletionItem.create(v, node, { kind: CompletionKind.Function }))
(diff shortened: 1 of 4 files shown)


packages/core/src/processor/completer/builtin.ts
```diff
@@ -211,15 +211,36 @@ export const resourceLocation: Completer<ResourceLocationNode> = (node, ctx) =>
...
+	const items = pool.map((v) => CompletionItem.create(v, node, { kind: CompletionKind.Function }))
+
+	if (node.options.category) {
+		const symbols = ctx.symbols.getVisibleSymbols(node.options.category, ctx.doc.uri)
+		const thisKey = Object.entries(symbols).flatMap(([key, symbol]) => {
+			if ((symbol.declaration?.[0] ?? symbol.definition?.[0])?.uri == ctx.doc.uri) {
+				return [key]
+			}
+			return []
+		})
```
NeunEinser, line 231: Couldn't you do sth like `ctx.symbols.query(ctx.doc, node.options.category, ...node.path).symbol` to get the symbol instead of iterating visible symbols manually?

misode: `node.path` would be the path of the resource location that you are typing. The problem is that I don't know the path of the symbol, that's the thing I'm looking for. I'm iterating over all symbols in the category and checking if that symbol is declared in this same file. Actually now that I type that out, I may also need to check that the contributor was `uri_binder`

NeunEinser: I see, we'd kind of need a `getDeclaredSymbols`, or context would need to know the current symbol (or symbol tree?) we are modifying.

If there is a better way to implement `getDeclaredSymbols` in symbols because of some access to some private member, that'd be nice, otherwise this is fine.

NeunEinser (approved):

packages/core/src/processor/completer/builtin.ts
misode: I have an idea on how to improve this in the future, I also found something similar in the mcdoc binder.

PR #1388, opened 2024-06-24
Revert changes to `%fallback`, WIP type generic in `itemStack`

MulverineX:
Related changes in vanilla-mcdoc: https://github.com/SpyglassMC/vanilla-mcdoc/commit/53a11fb03738fad0c68a89e98150aba30026c39b

### DIFF
--- a/packages/java-edition/src/mcfunction/checker/index.ts
+++ b/packages/java-edition/src/mcfunction/checker/index.ts
@@ -157,25 +157,24 @@ const itemStack: core.SyncChecker<ItemStackNode> = (node, ctx) => {
 		}
 		groupedComponents.get(componentId)!.push(pair.key)
 		if (pair.value) {
-			if (componentId === 'minecraft:custom_data') {
-				if (pair.value.type === 'nbt:string') {
-					// TODO: Maybe move this to the nbt package
-					const stringNBT = nbt.parser.compound(
-						new core.Source(pair.value.value, pair.value.valueMap),
-						ctx,
-					)
-					pair.value.children = [stringNBT]
-					core.AstNode.setParents(stringNBT)
-					// Because the runtime checker happens after binding, we need to manually call this
-					core.binder.fallbackSync(stringNBT, ctx)
-					core.checker.fallbackSync(stringNBT, ctx)
-					nbt.checker.index('mcdoc:custom_item_data', itemId)(stringNBT, ctx)
-				} else {
-					nbt.checker.index('mcdoc:custom_item_data', itemId)(pair.value, ctx)
-				}
-			} else {
-				nbt.checker.index('minecraft:data_component', componentId)(pair.value, ctx)
-			}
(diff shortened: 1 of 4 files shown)


packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -913,6 +913,10 @@ function simplifyIndexed<T>(
 	for (const index of typeDef.parallelIndices) {
 		let lookup: string[] = []
 		if (index.kind === 'static') {
+			if (index.value === '%fallback') {
+				values = child.fields.filter(f => f.kind === 'pair').map(f => f.type)
+				break
+			}
```
NeunEinser, line 919: I am already re-introducing this behavior in #1370, which also does some additional changes for performance, so the mcdoc/runtime changes should not be done in this PR to avoid conflicts.

packages/java-edition/src/mcfunction/checker/index.ts
```diff
@@ -157,25 +157,24 @@ const itemStack: core.SyncChecker<ItemStackNode> = (node, ctx) => {
...
-				} else {
-					nbt.checker.index('mcdoc:custom_item_data', itemId)(pair.value, ctx)
-				}
-			} else {
-				nbt.checker.index('minecraft:data_component', componentId)(pair.value, ctx)
-			}
+			const componentChecker = nbt.checker.index('minecraft:data_component', componentId, {
+				generics: [{
+					type: 'mcdoc:type_arg_block',
+					range: pair.value.range,
```
misode, line 163: What is this doing? Constructing a fake mcdoc node? Considering that we don't even have the `recommended` thing working, this can just be `nbt.checker.index('minecraft:data_component', componentId)`

packages/mcdoc/src/runtime/attribute/builtin.ts
```diff
@@ -138,4 +138,16 @@ export function registerBuiltinAttributes(meta: core.MetaRegistry) {
...
 		},
 	})
+
+	registerAttribute(meta, 'dispatcher_key', () => undefined, {
+		stringParser: (_, __, _ctx) => {
+			return (src, ctx) => {
+				return {
+					type: 'mcdoc:dispatcher_key',
+					range: core.Range.create(src.cursor, src.cursor),
+					value: Object.keys(_ctx.symbols.getVisibleSymbols('mcdoc/dispatcher')[core.string({})(src, ctx).value].members!),
```
misode, line 148: The way this is implemented right now probably means you'll get a list of all dispatchers. What you want is have this take a resource location as config, and query that specific dispatcher symbol.

misode (changes requested):

packages/mcdoc/src/runtime/checker/index.ts
MulverineX: Sounds good, I'll revert this

PR #1391, opened 2024-06-24
⚡ Add caching to mcdoc runtime

NeunEinser:
This caches the simplified version of McdocReferenceTypes and dispatcher indices, as long as no dynamic lookup was required during the resolution to a symplified typedef.


packages/mcdoc/src/binder/index.ts
```diff
@@ -113,6 +114,7 @@ namespace ModuleSymbolData {
 
 export interface TypeDefSymbolData {
 	typeDef: McdocType
+	simplifiedTypeDef?: SimplifiedMcdocType
 }
```
NeunEinser, line 118: Currently I am just amending the symbol, which means SimplifiedMcdocType is spreading again.
I think this should have the advantage that the cached value is removed for free when the symbol changes.

However, I am not sure how good this approach is. I didn't really figure out another good way to implement this though.

misode: This looks really dangerous to me, especially because `simplify` is not pure anymore and because simplified types also depend on runtime values.

NeunEinser: > especially because `simplify` is not pure anymore

If you mean the struct field thing from last PR, this is gone again in this PR, as I made field deduplication unnessary.

If you mean sth else, in what way is it not pure anymore?

> because simplified types also depend on runtime values.

The only part where runtime types are actually needed is for resolving dynamic dispatcher / indexed type indices. In these cases, I do not cache anything (dynamicData: true is returned)

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -1123,22 +1173,14 @@ function simplifyStruct<T>(typeDef: StructType, context: SimplifyContext<T>): Si
 		})
 
 		if (typeof key === 'string') {
-			literalFields.set(key, field)
-		} else if (key.kind === 'literal' && key.value.kind === 'string' && !key.attributes?.length) {
-			literalFields.set(key.value.value, field)
+			fields.push({
+				...field,
+				key: { kind: 'literal', value: { kind: 'string', value: key } },
```
misode, line 1178: How do you get away with not using a map for literal types here? How do you protect against adding a two fields with the same key?

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -791,30 +797,36 @@ export interface SimplifyValueNode<T> {
...
-		literalFields: Map<string, StructTypePairField>
-		complexFields: SimplifiedStructTypePairField[]
-	}
 	isMember?: boolean
 	typeArgs?: SimplifiedMcdocType[]
 	typeMapping?: { [path: string]: SimplifiedMcdocType }
 }
+export interface SimplifyResult<T extends SimplifiedMcdocType> {
+	typeDef: T
+	dynamicData?: boolean
```
misode, line 806: Not a fan of a special return type with a magic flag that enables/disables caching. This indicates to me that this is caching stuff on the wrong level. From what I can tell, if a single entity suddenly uses a dynamic index somewhere, it completely disables caching for the entity fallback case.

misode (commented):
I don't feel like this is the right caching approach

misode:
Also, it seems that loot entry, loot condition and loot function dispatching is broken, probably a lot more.

packages/mcdoc/src/runtime/checker/index.ts
NeunEinser: Basically, I realized that deduplicating here is mostly obsolete, as I already have a map of runtime keys in line 579. What will happen with multiple matching keys is, it first assigns the first one and later overwrites it with the second one.

The only case I needed to handle is complex keys, which is why I changed the code for those.

packages/mcdoc/src/runtime/checker/index.ts
NeunEinser: > From what I can tell, if a single entity suddenly uses a dynamic index somewhere, it completely disables caching for the entity fallback case.

Well, maybe yes? I am not sure you mean the correct thing here. Most of the result will still be cached. Caching something that requires a dynamic index is impossible (or you would need to cache the runtime value somehow, which I don't think would improve the performance, as the static index is still cached as I will explain below).

```rs
struct Llama {
    ...Breedable
    DecorItem?: InventoryItem
}

struct Breedable {
	...super::MobBase,
	/// Ticks until it stops searching for a mate.
	InLove?: int @ 0..,
	Age?: int,
	/// A value of age assigned to this mob when it grows up.
	/// Incremented when a baby mob is fed.
	ForcedAge?: int,
	/// Lower bits of the UUID of the player that caused this mob to breed.
	#[until="1.16"]
	LoveCauseLeast?: long,
	/// Upper bits of the UUID of the player that caused this mob to breed.
	#[until="1.16"]
	LoveCauseMost?: long,
	/// Player that caused this mob to breed.
	#[since="1.16"] #[uuid]
	LoveCause?: int[] @ 4,
}

struct Creeper {
	...super::MobBase,
	/// Whether it is being struck by lightning.
	powered?: boolean,
	/// Radius of the explosion.
	ExplosionRadius?: byte,
	/// Ticks until it explodes.
	Fuse?: short,
	/// Whether it was lit with flint and steel.
	ignited?: boolean,
}

struct MobBase {
	/// [main hand, offhand]
	HandItems?: [(InventoryItem | struct {})] @ 2,
	/// [feet, legs, body, head]
	ArmorItems?: [(InventoryItem | struct {})] @ 4,
}

dispatch minecraft:entity[creeper] to Creeper
dispatch minecraft:entity[llama] to Llama
```

Assuming we resolved the fallback case once, this will result in the following cached elements:
```md
- entity dispatcher
  - creeper
  - llama
- ::MobBase
- ::Breedable
- ::Creeper
- ::Llama
```

Now, let's say Llama was for some reason ammended with `...mincraft:spitting_behavior[[spitting_type]], spitting_type: string`

Now the following would be cached:
```md
- entity dispatcher
  - creeper
- ::MobBase
- ::Breedable
- ::Creeper
```

And when resolving the fallback dispatcher, llama would never hit the cache, never be cached and always resolved anew, however when resolving Llama, it would only do that, and not e.g. `::Breedable`.

Also note that this definition does not influence caching, since simplify is not deep:
```rs
struct {
    spitting_type: string,
    spitting_data: mincraft:spitting_behavior[[spitting_type]],
}
```

Also, this would be a hit after resolving fallback once with runtime value `{entity_id: creeper}`:
```rs
struct {
    entity_id: string
    entity_data: minecraft:entity[[entity_id]]
}
```
Basically, it would do the dynamic lookup, find creeper, then lookup creeper in the dispatcher and find the cached simplified value for creeper. This works, because the dynamic dispatcher `minecraft:entity[[entity_id]]` is not cached, but after resolving the dynamic indices, the result of the static members is cached.

> Not a fan of a special return type with a magic flag that enables/disables caching. This indicates to me that this is caching stuff on the wrong level.

Could you elaborate on this? I don't really agree, or understand that concern. Caching dynamic looksups is impossible by definition.

NeunEinser:
> Also, it seems that loot entry, loot condition and loot function dispatching is broken, probably a lot more.

Thanks for mentioning, this happened whenever you (even temporarily while typing) used an unknown value in a dispatcher.

This should be fixed by my last commit, however note that since this is amending symbols you may need to reset your cache for the fix to apply.

NeunEinser:
The only thing have which I find a little sub-optimal is attaching a SimplifiedMcdocType to the symbol directly like that. It feels like only the binder should write to symbols like that. This is just the easiest place I could come up with, and I don't know how I could introduce a cache in another way.

I do not understand / share  the concerns you have brought up with. I am mostly concerned about the place I am using as cache, while you seem to be concerned about in which place the caching happens.

I don't see any other sensible location to put the caching.

packages/mcdoc/src/runtime/checker/index.ts
misode: Hmm okay, I've slowly been realizing that maybe this is the best way to cache...

PR #1395, opened 2024-06-25
🐛 Allow downloader to use cache when offline

misode:
Two changes:
* Added an error hander on the nodejs downloader. Before this, the language server would crash when not connected to the network.
* ~~Allowed using the cache when fetching the checksum failed. This would allow using the cache both when offline and when the user is being rate limited by GitHub.~~

### DIFF
--- a/packages/core/src/common/externals/NodeJsExternals.ts
+++ b/packages/core/src/common/externals/NodeJsExternals.ts
@@ -37,6 +37,8 @@ class NodeJsExternalDownloader implements ExternalDownloader {
 				} else {
 					resolve(promisifyAsyncIterable(res, (chunks) => Buffer.concat(chunks)))
 				}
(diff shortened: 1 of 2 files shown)


SPGoding (commented):
If I understand the ten million `if` branches correctly, the original behavior when fetching the checksum failed was that we will try to 1) download the latest data and 2) fall back to the cache if the download failed. The only case where the downloader would return `undefined` is when both the network request and the cache loading failed. With the `checksum === undefined` change, step 1) would be skipped and we would only do 2) directly. Not sure if that's the desired behavior.

misode:
@SPGoding Okay then maybe only the first change is necessary to make this work offline. I will test and report back.

misode:
Can confirm it works as you described!

PR #1402, opened 2024-06-25
🐛 Deduplicate definitions in the checker tree

### LINKED ISSUE #1401: Redundant hover information when displaying union types, especially visible with entities
image(16)
This us caused by all entities contributing their own duplicate definitions for both value and key.

We should deduplicate these unions, potentially even reducing the definitions that need to be checked for children.

NeunEinser:
Fixes #1401

This deduplicates definitions when adding them to the checker tree. This is done as early as possible which has a few advantages performance-wise:

## Complexity of deduplicating itself
By definition, deduplication means complexity O(n^2) which isn't great, as each definition needs to be checked against each other definition.

This approach reduces this to be always effectively O(n*m), where m is the number of _unique_ definitions, and n is the number of _total_defintions.

This means each definition is guranteed to be checked only against other uniquely different definitions, since we only collect definitions that differ in the first place, and never write any duplicates

So this is only bad compexity wise if we have a definition that means that a child for some reason gets a ton of valid definitions which are all different from each other, which really shouldn't happen.

## Performance gains
On types that have a lot of duplicate definitions for some nodes (one example is a common base type being spread over all types of a dispatcher, while using the fallback case; in this case all children of the struct which appear in the common base type are duplicated to all dispatched values), this means:
- simplify() is only called once per unique definition.
- The whole checking logic is also only executed once per unique definition

I expect some performance improvements from this for such types, one example would be the entity fallback which is sometimes used in commands.

## The fix in action
In particular, #1401 is fixed:
image

There however still seems to be a bug with error condensing here:
image
I'll take a look at the remaining issue and see how easy it is to fix.

### DIFF
--- a/packages/mcdoc/src/runtime/checker/error.ts
+++ b/packages/mcdoc/src/runtime/checker/error.ts
@@ -4,7 +4,9 @@ import { arrayToMessage, localeQuote, localize } from '@spyglassmc/locales'
 import { RangeKind } from '../../node/index.js'
 import { McdocType, NumericRange } from '../../type/index.js'
 import type {
+	CheckerTreeDefinitionGroupNode,
 	CheckerTreeDefinitionNode,
+	CheckerTreeNode,
 	ErrorReporter,
 	RuntimeNode,
 	SimplifiedMcdocTypeNoUnion,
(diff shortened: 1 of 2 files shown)


packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -271,7 +273,23 @@ export function typeDefinition<T>(
...
+											: McdocType.equals(d.keyDefinition, childDef.keyType))
+										&& McdocType.equals(d.originalTypeDef, childDef.type),
+								)
+							: -1
+
 						for (const childValue of child.possibleValues) {
+							if (existingDefIndex >= 0) {
+								childValue.definitionsByParent[existingDefIndex].parents.push(def)
+								def.children.push(childValue.definitionsByParent[existingDefIndex])
+								continue
```
TheAfroOfDoom, line 291: i think i mostly understand what this is doing. would this refactor be correct (naming-wise?)

```suggestion
							if (existingDefIndex >= 0) {
								const existingDef = childValue.definitionsByParent[existingDefIndex]
								existingDef.parents.push(def)
								def.children.push(existingDef)
								continue
```

packages/mcdoc/src/runtime/checker/error.ts
```diff
@@ -112,7 +114,101 @@ export interface ErrorCondensingDefinition<T> {
...
-export function condenseErrorsAndFilterSiblings<T>(definitions: ErrorCondensingDefinition<T>[]): {
+export function condenseAndPropagate<T>(
+	definitionGroup: CheckerTreeDefinitionGroupNode<T>,
+	definitionErrors: ErrorCondensingDefinition<T>[],
+) {
+	const queue = [{ node: definitionGroup, errorsOnLayer: definitionErrors, depth: 0 }]
+	while (queue.length) {
+		const { node, errorsOnLayer, depth } = queue.shift()!
+
+		const stillValidDefintions: CheckerTreeDefinitionNode<T>[] = []
```
TheAfroOfDoom, line 125:
```suggestion
		const stillValidDefinitions: CheckerTreeDefinitionNode<T>[] = []
```

TheAfroOfDoom (approved):
i understand what this is doing high-level enough. good stuff 😄

NeunEinser:
> There however still seems to be a bug with error condensing here: image I think this is caused by the root itself being a union, so it reports all errors from all members in this case. This is a case that was not really considered when implementing the condensing logic, but might just be another call at the very end.
> 
> I think I would like to improve error condensing separetely though, especially since this seems less severe compared to the issue being fixed here.

Actually, that doesn't really make sense, root is expected to always just have one definition group with all definitions, so it really should be condensed correctly there. And condensing does happen since error reporting only looks at the root and would not find anything otherwise.

packages/mcdoc/src/runtime/checker/index.ts
NeunEinser: Yeah, since it was only used in two places I didn't really see the need to name a constant ¯\_(ツ)_/¯

NeunEinser:
Fixed the bug with duplicate error messages (and added a test case for it)

packages/mcdoc/src/runtime/checker/index.ts
TheAfroOfDoom: eh, the expression is so long and the context not immediately clear enough that I think the named comst helps new readers quite a bit

packages/mcdoc/test/runtime/checker.spec.ts
```diff
@@ -45,6 +45,35 @@ describe('mcdoc runtime checker', () => {
 		values: [{}, { text: 'something' }, { selector: 20 }, { text: 'foo', selector: 40 }, {
 			selector: [1],
 		}],
+	}, {
+		name:
+			'( struct { one?: string, two?: int, three?: boolean } | struct { one?: text, four?: int, five?: boolean })',
```
misode, line 50: `one?: text` -> `one?: string`

packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -271,7 +273,24 @@ export function typeDefinition<T>(
 							continue
 						}
 						const child = childNodes[i]
+						const existingDefIndex = child.possibleValues.length > 0
+							? child.possibleValues[0].definitionsByParent
+								.findIndex(
```
misode, line 278: You could use `find` here instead

misode (commented):
Looks pretty good!

I found a small regression: empty unions are no longer reported as "expected nothing". Example command:
```
data merge entity @s {id:"test"}
```
It seems likely this is an issue with the vanilla-mcdoc, because `EntityBase` no longer defines an `id` field for the player to overwrite, but maybe it is simple enough to fix.

> This is done as early as possible

Is it not possible to deduplicate equal union members during simplify? Or is the advantage then that you lose parent type information?

NeunEinser:
> Is it not possible to deduplicate equal union members during simplify? Or is the problem then that you lose parent type information?

No, this would just achieve sth different.

This PR is about valid definitions from different parents. Simplify sees valid definitions from the same parent, so those are not deduplicated at all.

Your suggestion would deduplicate this: `struct { one: (string | string) }`, but `one` would be two definition groups with two separate McdocTypes which are handled by two separate `simplify` calls before this PR: `(struct { one: string} | struct { one: string, two: int} )`.

This PR does the opposite: It handles the case deduplicating in simplify cannot handle, but it doesn't deduplicate duplicated definitions from a single original non-simplified type def. So, the validDefinitions array within one definition group may contain duplicates. 

Sth that neither of the approaches gandles is different definition groups could also share types of the typedef is different, for example `(struct { foo: (string | int)} | struct { foo: (string | boolean)})`, now there would be two definition groups, one with string | int, the other with string | boolean, so `string` is duplicated between the two definition groups and would be reported twice. In this example, the key would also still be reported as (foo | foo). The same would also happen here: `type StringType = string; (struct { foo: string} | struct { foo: StringType })`, because the typedefs are cimpared before simplify and before references are resolved.

So there are still some cases were it would happen, and to fix those, you just need to do the naive thing of deduplicating when attaching type info.

NeunEinser:
> Looks pretty good!
> 
> I found a small regression: empty unions are no longer reported as "expected nothing". Example command:
> 
> ```
> data merge entity @s {id:"test"}
> ```

I added a test case for empty union, but it is behaving as expected?

misode:
> but it is behaving as expected?

Okay weird it is indeed working fine. I'm going to blame caching 😄

misode (approved):

PR #1405, opened 2024-06-26
🐛 Fix data append and prepend using list definition instead of list item for type checking

### LINKED ISSUE #1403: Incorrect type indexing with `data modify ... insert|prepend|append`
```
data modify block ~ ~ ~ Items append value {id:"book",count:1}
```

NeunEinser:
Fixes #1403

With this, the list item definition is used instead for append and prepend.

We probably want to verify that the item returned by the path is actually a list. The type checking for the Nbt node is completely disabled when no list-like is found at the index.

Similarly, there is currently no error when trying to use data modify merge on anything else than a compound.

### DIFF
--- a/packages/java-edition/src/mcfunction/checker/index.ts
+++ b/packages/java-edition/src/mcfunction/checker/index.ts
@@ -258,8 +258,6 @@ function getListLikeChild(typeDef: mcdoc.McdocType): mcdoc.McdocType | undefined
 			return { kind: 'int' }
 		case 'long_array':
 			return { kind: 'long' }
-		case 'tuple':
-			return { kind: 'union', members: typeDef.items }
 		case 'union':
 			const members = typeDef.members
 				.map(m => getListLikeChild(m))
--- a/packages/java-edition/src/mcfunction/checker/index.ts
+++ b/packages/java-edition/src/mcfunction/checker/index.ts
@@ -202,12 +202,12 @@ function nbtChecker(
 		const tag = node.children[0]
 		if (indexedBy) {
 			if (NbtPathNode.is(indexedBy)) {
-				let typeDef: mcdoc.McdocType | undefined = indexedBy.children[0].endTypeDef
-				if (typeDef && node.properties.isListIndex) {
-					typeDef = getListLikeChild(typeDef)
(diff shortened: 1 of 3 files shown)


NeunEinser:
Tuples don't really make sense to be appended or prepended, they should always have a fixed size. The only case I suppose would be sth like `([string] | [string, int])` where you could technically append an int and it would still be valid.

I think it's fine to just not support this though.

packages/java-edition/src/mcfunction/checker/index.ts
```diff
@@ -242,6 +248,33 @@ function nbtChecker(
 	}
 }
 
+function getListLikeChild(typeDef: mcdoc.McdocType): mcdoc.McdocType | undefined {
```
misode, line 251: Should this be changed to take a `SimplifiedMcdocType`?

misode (approved):
Clean solution, I like it!

packages/java-edition/src/mcfunction/checker/index.ts
```diff
@@ -201,8 +201,14 @@ function nbtChecker(
...
 		if (indexedBy) {
-			if (NbtPathNode.is(indexedBy) && indexedBy.children[0].endTypeDef) {
-				nbt.checker.typeDefinition(indexedBy.children[0].endTypeDef)(tag, ctx)
+			if (NbtPathNode.is(indexedBy)) {
+				let typeDef: mcdoc.McdocType | undefined = indexedBy.children[0].endTypeDef
+				if (typeDef && node.properties.isListIndex) {
+					typeDef = getListLikeChild(typeDef)
+				}
+				if (typeDef) {
+					nbt.checker.typeDefinition(typeDef)(tag, ctx)
```
misode, line 210: This was a bug before this PR too, but this call should pass `isMerge`

packages/java-edition/src/mcfunction/checker/index.ts
NeunEinser: The return value no, because the item of a list is not simplified. The parameter could be though.

SPGoding:
Nice

Issue #1415, opened 2024-06-29
Ability to Toggle Minecraft Namespace Autocompletion


catter1:
When typing in registry names (items, features, tags, etc), Spyglass will autocomplete it without the `minecraft:` namespace before it, unless explicitly typed.

It would be nice to have an extension setting to force the autocompletion of the namespace where applicable. I know it's a niche feature request, but hey, I like having everything look the same!

catter1:
Note: In instances such as referencing components in `data` commands, the namespace is *required*. ([see this discord message](https://discord.com/channels/666020457568403505/1255970681930252320/1256542689248149514)). So at the very least, it should be autocompleted by default where it is required. I wrote this issue when thinking of JSON, where it is optional.

InQuognito:
lol i’m the same way, i never use the auto complete for the minecraft registry because i want the namespace. i second this

SPGoding:
It's already implemented. Set `lint.idOmitDefaultNamespace` to `false` in `spyglassmc.json`.

PR #1416, opened 2024-06-29
🐛 Fix auto complete suggesting namespace-less entries with isPredicate

NeunEinser:
This was mentioned on discord while discussing #1415 

Json never has a predicate that requires specifying the namespace, so JSON remains unchanged

For SNBT, I save whether checking was using the isPredicate ruleset on the NBT node when attaching the mcdoc typeinfo.

For NBT paths, this is always true, similar to how NBT paths are always checked with the isPredicate ruleset.

There is a new mcdoc completer context now which extends the core completer context with an optional requireCannoical parameter.

### DIFF
--- a/packages/mcdoc/src/runtime/completer/index.ts
+++ b/packages/mcdoc/src/runtime/completer/index.ts
@@ -6,9 +6,13 @@ import type { SimplifiedEnum, SimplifiedMcdocType } from '../checker/index.js'
 
 export type SimpleCompletionField = { key: string; field: core.DeepReadonly<StructTypePairField> }
 
+export interface McdocCompleterContext extends core.CompleterContext {
+	requireCannonical?: boolean
+}
+
 export function getFields(
 	typeDef: core.DeepReadonly<SimplifiedMcdocType>,
-	ctx: core.CompleterContext,
+	ctx: McdocCompleterContext,
 ): SimpleCompletionField[] {
 	switch (typeDef.kind) {
 		case 'union':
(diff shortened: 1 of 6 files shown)


packages/mcdoc/src/runtime/completer/index.ts
```diff
@@ -6,9 +6,13 @@ import type { SimplifiedEnum, SimplifiedMcdocType } from '../checker/index.js'
 
 export type SimpleCompletionField = { key: string; field: core.DeepReadonly<StructTypePairField> }
 
+export interface McdocCompleterContext extends core.CompleterContext {
+	requireCannonical?: boolean
```
misode, line 10: Should be `requireCanonical`

packages/nbt/src/node/index.ts
```diff
@@ -3,6 +3,7 @@ import type * as mcdoc from '@spyglassmc/mcdoc'
 
 interface NbtBaseNode {
 	typeDef?: mcdoc.runtime.checker.SimplifiedMcdocType
+	requireCannonical?: boolean
```
misode, line 6: Same here

misode (commented):
Looks good otherwise!

packages/mcdoc/src/runtime/completer/index.ts
NeunEinser: Huh, TIL.

misode (approved):

PR #1421, opened 2024-06-29
✨ Create Zed Editor extension

MulverineX:


TheAfroOfDoom:
do Zed extensions _need_ to be written in rust? if not, seems weird to add a new language to the 99% TS repo :L

MulverineX:
Multi-language mono-repos are not as uncommon as you'd think, and I'm not sure what Zig uses under the hood as a communication protocol with extensions, but there isn't API implementations available for any other language, and rolling our own would be much more work.

MulverineX:
Progress update: https://github.com/MulverineX/tree-sitter-mcfunction (this replaces the textmate syntax-mcfunction that we depend on in VSCode)

PR #1423, opened 2024-06-30
🔧 Add publish vscode extension step to release workflow

misode:
Very much untested, taken mostly from <https://github.com/SpyglassMC/vscode-datapack/blob/legacy-v3/.github/workflows/release.yml>


SPGoding (commented):
I think we can move the release commands to the `release` script in `packages/vscode-extension`, so they all get run by `npm run release` at root? The commands and environmental variables look good to me.

PR #1428, opened 2024-06-30
🐛 Fix crash when fileModified before dependencyFiles is set

misode:
We got a report on discord where the stacktrace indicated that `isOnlyWatched` was being called from the `on fileModified` handler while `#dependencyFiles` was still undefined. The `fileModified` handler is attached in the contructor, while `#dependencyFiles` is only set during the `ready` phase, so this is not unexpected.

Is it better practice to just initialize these two as empty sets?


misode:
> Patch works, and it's no longer crashing 🎉

Issue #1212, opened 2024-05-28
Changing the pack format does not invalidate cache and re-initialize the project


misode:
> What I want for #1212 is that when you change the pack format either a warning shows in pack.mcmeta with link to reload the project, or it automatically reloads the project.

Issue #1071, opened 2023-08-05
Support multi-version packs


SPGoding:
To keep the mental model easy we can construct a separate symbol table for each version that the data pack claims to support. Common symbols may be shared across the symbol tables to save memory (why isn't the symbol table immutable who the heck wrote this piece of hot garbage). We can then attempt to validate the entire data pack for each version although that could be a performance nightmare, especially when code in the common part of all versions is changed. Not sure if there's any way around it though.

NeunEinser:
Possibly the parser could start with some supported version and keep track of versions were there were no changes to the thing it's parsing, based on mcdoc (and have the version range that satisfies everything in a list).

Then, only re-parse with another version, if that's not added to the supported range by the first pass.

Bowswa:
I see the potential issue with performance, so i have a couple suggestions to attach:
Problem #1 - at least to my knowledge, you cannot specify more than 2 specific pack formats, it's always parsed as a range of them. 
Solution #1a - Flag snapshot overlays as "unsupported", but not a warning to annoy us all
Solution #1b - Initial parsing parses the main data folder, then each overlay as it gets activated going up the versions(overlay 44 would load & be parsed, then overlay 55 would load on TOP and be parsed for example, and personally i name my overlays based on pack format they support). Then only re-parse each file as it is changed, and everything else is cached.

Problem #2 - parsing too many overlays being laggy
Solution #2 - while this is an obscure one, when parsing more than say 7 overlays, instead of parsing on update, it either parses like once per minute, or can be manually told to re-parse.

PR #1469, opened 2024-07-03
✨ Add pack format mcdoc attribute

misode:
image
image
image

### DIFF
--- a/packages/java-edition/src/index.ts
+++ b/packages/java-edition/src/index.ts
@@ -3,7 +3,7 @@ import * as json from '@spyglassmc/json'
 import * as mcdoc from '@spyglassmc/mcdoc'
 import * as nbt from '@spyglassmc/nbt'
 import { uriBinder } from './binder/index.js'
-import type { McmetaSummary } from './dependency/index.js'
+import type { McmetaSummary, McmetaVersion } from './dependency/index.js'
 import {
 	getMcmetaSummary,
 	getVanillaDatapack,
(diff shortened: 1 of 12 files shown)


packages/java-edition/src/index.ts
```diff
@@ -156,6 +157,44 @@ export const initialize: core.ProjectInitializer = async (ctx) => {
...
+						node,
+						core.ErrorSeverity.Warning,
+					)
+				} else if (targetVersion.id !== release) {
+					ctx.err.report(
+						localize('java-edition.pack-format.not-loaded', target, release),
+						node,
+						core.ErrorSeverity.Warning,
+					)
+				}
```
TheAfroOfDoom, line 186: instead of raising a warning in a file most people likely won't check, why don't we just forcibly delete the cache here? (or restart spyglass, whatever is needed to re-parse the `pack.mcmeta` and datapack contents)

misode: A user could have multiple pack.mcmeta files, or they could have an explicit `gameVersion` config set, or they may want to change the pack_format to match the currently loaded version instead.

Automatically restarting vscode or loading the correct version is out of scope for this PR, and will be handled when fixing #1071 and #1212

misode:
I'm merging this, because I want to start with #1269 and want to avoid conflicts. I know the two new attribute handlers are not the cleanest, but it can be refactored later.

PR #1472, opened 2024-07-04
🐛 Fix nbt paths ending with an enum type

### LINKED ISSUE #1437: "data modify block" commands affecting sign front_text.color and back_text.color show incorrect errors
The following valid `data modify block` commands error out on the token `color`:

```mcfunction
data modify block -54 -24 9 front_text.color set value "light_gray"
data modify block -54 -24 9 back_text.color set value "light_gray"
```

image
image

The same behavior does not occur for `data merge block` equivalents:
```mcfunction
data merge block -54 -24 9 {front_text:{color:"light_gray"}}
data merge block -54 -24 9 {back_text:{color:"light_gray"}}
```
image

misode:
- Fixes #1437 

Because NBT paths uses `unsafe` as their leaf type, this would cause a type mismatch when the nbt paths ended with an enum (or literal)

### DIFF
--- a/packages/mcdoc/src/runtime/checker/index.ts
+++ b/packages/mcdoc/src/runtime/checker/index.ts
@@ -465,11 +465,15 @@ function checkShallowly<T>(
 	}
 
 	if (
-		(typeDef.kind === 'literal'
+		(typeDef.kind !== 'any' && typeDef.kind !== 'unsafe'
+			&& simplifiedInferred.kind !== 'unsafe'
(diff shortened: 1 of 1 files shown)


NeunEinser (commented):
There are a couple of these checks now (one in line 444 too).

Would it make sense to just return early in these cases instead of checking this in 3 different places?

NeunEinser (commented):
(approved, was late to review after your changes)

PR #1473, opened 2024-07-04
✨ Add advancement criterion checking and suggestions

### LINKED ISSUE #883: Provide suggestions for advancement criteria in commands
Relates to #835.

misode:
- Closes #883


misode:
To enter the criterion symbols, I've used the `advancement` category with the advancement ID as the `parentPath`. However I'm not sure this is the right call, as for example the hover information will now say `(advancement) iron_pickaxe`. Maybe this is a problem that needs to be solved in `Service#getHover`. I thought about using a separate `advancement_criterion` category, but then the problem was that it would throw errors that the parent map didn't exist. @SPGoding

SPGoding:
I think fixing the `getHover()` function would be a better move here.

SPGoding:
Coming back to this, setting `subcategory` on the child symbol to `criterion` should result in a hover like `(advancement/criterion) iron_pickaxe`. I think that'd be good enough?

PR #1485, opened 2024-07-07
✨ Add command too long check

### LINKED ISSUE #1165: Validate the maximum length of a command in a function
In 24w06a a limit was introduced for the length of commands in a function: 2000000 characters. This should be validated.

misode:
- Closes #1165


misode:
I think this currently checks the length of the command node including line continuations, so it might warn sooner then necessary, but it's better than nothing. Moving the check to inside the parser would be a bit more difficult

NeunEinser:
Note that click events used in tellraw have a way smaller length. They use the maximum length you can use in chat.

I think books behave the same, but signs don't

PR #1486, opened 2024-07-07
🔥 Remove IndexMap.merge

### LINKED ISSUE #1296: Index map issue with nested strings

misode:
- [x] Removed `IndexMap.merge`, it's not needed and was broken anyways
- [x] Fixed #1296

This fix likely makes the string parser a lot slower, because it needs to compute the outer cursor for every character. There is likely a more efficient way of combining the relevant index map pairs of the src indexMap, but that seems tricky to implement.


misode:
image

TheAfroOfDoom (approved):

TheAfroOfDoom:
> because it needs to compute the outer cursor for every character.

this was also my expected solution, given a cleaner IndexMap.merge method not yet existing

PR #1487, opened 2024-07-07
♻️ Use bigint for long types in mcdoc

NeunEinser:


packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -498,7 +500,8 @@ function checkShallowly<T>(
 			if (
 				typeDef.valueRange
 				&& simplifiedInferred.kind === 'literal'
-				&& typeof simplifiedInferred.value.value === 'number'
+				&& simplifiedInferred.value.kind !== 'string'
+				&& simplifiedInferred.value.kind !== 'boolean'
```
misode, line 504: There are probably a few other places, maybe in attributes, where this pattern was used

packages/mcdoc/src/binder/index.ts
```diff
@@ -1001,7 +1009,7 @@ function convertLiteralValue(node: LiteralTypeValueNode, ctx: McdocBinderContext
 			kind: convertLiteralNumberSuffix(suffix, ctx)
 				?? (value.type === 'integer' ? 'int' : 'double'),
 			value: value.value,
-		}
+		} as LiteralValue
```
misode, line 1012: This PR adds a couple of these unsafe casts, if you find a way to get rid of them that would be nice

misode (commented):
Some of these unsafe casts make me a bit nervous, but I don't immediately see a clean way to get rid of them

packages/mcdoc/src/runtime/checker/index.ts
NeunEinser: Pretty sure that's untrue. This one I cought because it caused a build error, so I don't see why this wouldn't happen in other places.

packages/mcdoc/src/binder/index.ts
NeunEinser: In some places, you can do

```ts
return kind === 'long'
  ? { kind, value }
   : { kind, value }
```

to satisfy the type checker, but it's also a bit weird ¯\\\_(ツ)\_/¯

misode:
I had to revert this, because this caused issues when saving the symbols in the cache file. I don't see a button to reopen the PR, so you might need to open a new PR when we have a fix for this.
```
[Error - 1:04:51 AM] [CacheService#save] path = file:///<CACHE>/symbols/d354ae857b06ee874f712c28b13436746cebac42.json.gz TypeError: Do not know how to serialize a BigInt
    at JSON.stringify (<anonymous>)
    at Object.writeGzippedJson (<SOURCE>\packages\core\src\service\fileUtil.ts:321:49)
    at CacheService.save (<SOURCE>\packages\core\src\service\CacheService.ts:234:19)
```

PR #1491, opened 2024-07-08
✨ Validate leading slash in command attribute

### LINKED ISSUE #818: commands in JSON text components require a leading '/', even on a sign to be evaluated
Dph+ will ignore any command specified in a click_event of a JSON text component, unless it includes a leading slash.

However, on signs, when using the `run_command` action, in contrast to everywhere else, the command does not run fully in the player's scope and thus does not require a leading slash in this case to work in-game.

It could also be nice to mark an error when there is no leading slash in other locations, though maybe sending a chat message is wanted in some cases.

misode:


packages/java-edition/src/mcfunction/mcdocAttributes.ts
```diff
@@ -32,11 +48,19 @@ const scoreHolderValidator = validator.alternatives<ScoreHolderConfig>(
...
 export function registerMcdocAttributes(meta: core.MetaRegistry, rootTreeNode: mcf.RootTreeNode) {
-	mcdoc.runtime.registerAttribute(meta, 'command', () => undefined, {
-		// TODO: validate slash
+	mcdoc.runtime.registerAttribute(meta, 'command', commandValidator, {
 		// TODO: fix completer inside commands
-		stringParser: () => {
-			return mcf.command(rootTreeNode, parser.argument)
+		stringParser: ({ slash, empty }) => {
+			return (src, ctx) => {
+				if ((empty && !src.canRead()) || (slash === 'chat' && src.peek() === '/')) {
```
MulverineX, line 55: Is this supposed to be `src.peek() !== '/'`?

misode: It looks like it yeah, good catch!

SPGoding (approved):

MulverineX (approved):

packages/java-edition/src/mcfunction/mcdocAttributes.ts
```diff
@@ -32,11 +46,20 @@ const scoreHolderValidator = validator.alternatives<ScoreHolderConfig>(
...
-			return mcf.command(rootTreeNode, parser.argument)
+		stringParser: ({ slash, max_length, empty }) => {
+			return (src, ctx) => {
+				if ((empty && !src.canRead()) || (slash === 'chat' && src.peek() !== '/')) {
+					return core.string({
+						unquotable: { blockList: new Set(), allowEmpty: true },
+					})(src, ctx)
+				}
+				return mcf.command(rootTreeNode, parser.argument, {
+					slash: slash === 'chat' ? 'allowed' : slash,
```
NeunEinser, line 59: If it's chat as in a /tellraw component, a slash is actually required (without the slash, it used to just send a chat message in your name, but I think since the chat signing thing, it's just disallowed).

On signs however, a slash is optional.

If the aim is not to 100% solve these cases, the linked issue as wrong, as it describes a v3 bug where DPH+ would show an error when omitting the slash on a sign.

packages/mcdoc/src/runtime/attribute/builtin.ts
```diff
@@ -150,6 +169,11 @@ export function registerBuiltinAttributes(meta: core.MetaRegistry) {
 			return resourceLocation
 		},
 	})
+	registerAttribute(meta, 'integer', integerValidator, {
+		stringParser: (config) => {
+			return core.integer({ pattern: /^-?\d+$/, min: config.min, max: config.max })
+		},
+	})
```
NeunEinser, line 176: Cursed Mojang syntax

NeunEinser (approved):

packages/java-edition/src/mcfunction/mcdocAttributes.ts
misode: Hmm so on signs in the `run_command` action, it will run the command when you don't start with a slash?

PR #1494, opened 2024-07-08
📝 Write user-friendly mcdoc guide

misode:


docs/user/mcdoc/attributes.adoc
MulverineX, line 1: imo, the ordering here shouldn't be alphabetical. we should group things like since/until, and attributes that are more useful for datapack mcdoc types should be pushed to the top

MulverineX (approved):
We can merge like this, but if you're willing to reorder that'd be great

PR #1495, opened 2024-07-08
🐛 Fix mapped types being cached and reused with different mappings

NeunEinser:
Disclaimer: I made this on phone and didn't test it.

You may need to clear all cached data to see any effect.


misode (approved):
This seems to fix both the particle issue that was reported and the issues reported on Discord

PR #1505, opened 2024-07-09
✨ Add "Show Output" command

misode:
Because it is annoying to tell people how to find the output logs

### DIFF
--- a/docs/user/commands.adoc
+++ b/docs/user/commands.adoc
@@ -8,6 +8,10 @@
 
 Show the cache folder of Spyglass in the platform-specific explorer program.
 
+=== `spyglassmc.showOutput`
+
+Show the output panel with the logs of the Spyglass Language Server.
+
 === `spyglassmc.resetProjectCache`
 
 Reset the symbols cache for the current project.
--- a/packages/vscode-extension/src/extension.mts
+++ b/packages/vscode-extension/src/extension.mts
@@ -119,7 +119,7 @@ export async function activate(context: vsc.ExtensionContext) {
 
 		if (customCapabilities?.showCacheRoot) {
 			context.subscriptions.push(
-				vsc.commands.registerCommand('spyglassmc.showCacheRoot', async (): Promise<void> => {
+				vsc.commands.registerCommand('spyglassmc.showCacheRoot', async () => {
 					try {
 						await client.sendRequest('spyglassmc/showCacheRoot')
 					} catch (e) {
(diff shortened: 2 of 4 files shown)


packages/vscode-extension/src/extension.mts
```diff
@@ -119,7 +119,7 @@ export async function activate(context: vsc.ExtensionContext) {
 
 		if (customCapabilities?.showCacheRoot) {
 			context.subscriptions.push(
-				vsc.commands.registerCommand('spyglassmc.showCacheRoot', async (): Promise<void> => {
+				vsc.commands.registerCommand('spyglassmc.showCacheRoot', async () => {
```
TheAfroOfDoom, line 122: guessing the `Promise<void>` is implied here?

packages/vscode-extension/src/extension.mts
```diff
@@ -129,6 +129,16 @@ export async function activate(context: vsc.ExtensionContext) {
 			)
 		}
 
+		context.subscriptions.push(
+			vsc.commands.registerCommand('spyglassmc.showOutput', () => {
```
TheAfroOfDoom, line 133: it'd be really nice if all these commands were refactored into an [enum](https://www.typescriptlang.org/docs/handbook/enums.html) so we aren't copy-pasting these strings.

```ts
enum VscodeCommand {
	ShowCacheRoot = 'showCacheRoot',
	ShowOutput = 'showOutput',
	// ...
}
```

can i make a PR that does this refactor?

TheAfroOfDoom: on second thought this may not be that useful since we need these strings in `packages/vscode-extension/package.json`. so we can't actually reference these enums.

docs/user/commands.adoc
```diff
@@ -8,6 +8,10 @@
 
 Show the cache folder of Spyglass in the platform-specific explorer program.
 
+=== `spyglassmc.showOutput`
```
TheAfroOfDoom, line 11: this file not being alphabetical is killing me 😭

TheAfroOfDoom (approved):

packages/vscode-extension/src/extension.mts
misode: Yes. The other commands also don't have it

docs/user/commands.adoc
MulverineX: https://github.com/SpyglassMC/Spyglass/issues/1553

MulverineX (approved):

PR #1506, opened 2024-07-10
✨ Add gameVersion to vscode settings

misode:
### Future work
- Restart/notify the language server when the setting changes

### DIFF
--- a/packages/vscode-extension/package.nls.json
+++ b/packages/vscode-extension/package.nls.json
@@ -4,5 +4,6 @@
     "spyglassmc.commands.showCacheRoot.title": "Open Cache Folder",
     "spyglassmc.description": "Heavy language features for Minecraft: Java Edition data pack files",
     "spyglassmc.displayName": "Datapack Helper Plus by Spyglass",
+    "spyglassmc.env.gameVersion": "The Minecraft version to use. When set to auto, the version is inferred from the `pack.mcmeta`. Other versions can be specified in a `spyglass.json` config file.",
     "spyglassmc.trace.server": "Traces the communication between VS Code and the Spyglass Language Server."
 }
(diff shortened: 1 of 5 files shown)


packages/vscode-extension/package.nls.json
```diff
@@ -4,5 +4,6 @@
     "spyglassmc.commands.showCacheRoot.title": "Open Cache Folder",
     "spyglassmc.description": "Heavy language features for Minecraft: Java Edition data pack files",
     "spyglassmc.displayName": "Datapack Helper Plus by Spyglass",
+    "spyglassmc.env.gameVersion": "The Minecraft version to use. When set to auto, the version is inferred from the `pack.mcmeta`. Other versions can be specified in a `spyglass.json` config file.",
```
TheAfroOfDoom, line 7: what does `nls` stand for in this file's name?

packages/vscode-extension/src/extension.mts
```diff
@@ -46,8 +46,11 @@ export async function activate(context: vsc.ExtensionContext) {
 		{ language: 'json', pattern: '**/data/*/*/**/*.json' },
 	]
 
+	const gameVersion = vsc.workspace.getConfiguration('spyglassmc.env').get('gameVersion')
+
 	const initializationOptions: server.CustomInitializationOptions = {
 		inDevelopmentMode: context.extensionMode === vsc.ExtensionMode.Development,
+		gameVersion: typeof gameVersion === 'string' ? gameVersion : undefined,
```
TheAfroOfDoom, line 53: my thought was to maybe raise an error if `typeof gameVersion === 'string'` was false, but i guess vscode will just display an error in the settings.json for us instead?

misode: Yes, it will show a warning in the settings file, so the extension should gracefully ignore it.
image

MulverineX (approved):

packages/vscode-extension/package.nls.json
SPGoding: "National Language Support" or "Native Language Support", VS Code's choice of word for i18n. This file is picked up by VS Code automatically for localization.

PR #1523, opened 2024-07-14
🐛 Enable `usePolling` on Windows to fix file-lock

### LINKED ISSUE #1414: Cannot rename folder
> [!NOTE]
> This issue is caused on Windows due to the file watcher we are using. **You can try the following workaround**:
> 1. Create a file `spyglass.json` in the workspace root
> 2. Paste the following contents inside:
> ```json
> {
>   "env": {
>     "useFilePolling": true
>   }
> }
> ```
> 3. Reload vscode, for example by running the `Developer: Reload Window` command after pressing Ctrl+Shift+P

https://github.com/SpyglassMC/Spyglass/assets/154045603/95f0fbd5-9ede-4684-9088-90f21432d903

TheAfroOfDoom:
# Summary

- closes #1414 

There hasn't been a noticeable performance drop in my (fairly large) project so I think this solution is more than good enough 🤷 

(if anyone wants to do some performance checks on this in voice chat with me I'd be more than happy to. not entirely sure what to check other than speed at which a file is re-parsed upon changing)

## Preview

| before | after |
|-|-|
|before renaming|after renaming|

### DIFF
--- a/packages/core/src/common/externals/NodeJsExternals.ts
+++ b/packages/core/src/common/externals/NodeJsExternals.ts
@@ -117,9 +117,9 @@ export const NodeJsExternals: Externals = {
 		unlink(location) {
 			return fsp.unlink(toFsPathLike(location))
 		},
-		watch(locations) {
+		watch(locations, { usePolling = false } = {}) {
 			return new ChokidarWatcherWrapper(
-				chokidar.watch(locations.map(toPath), { usePolling: process.platform === 'win32' }),
+				chokidar.watch(locations.map(toPath), { usePolling }),
 			)
 		},
 		writeFile(location, data, options) {
--- a/packages/core/src/common/externals/NodeJsExternals.ts
+++ b/packages/core/src/common/externals/NodeJsExternals.ts
@@ -118,7 +118,9 @@ export const NodeJsExternals: Externals = {
 			return fsp.unlink(toFsPathLike(location))
 		},
 		watch(locations) {
-			return new ChokidarWatcherWrapper(chokidar.watch(locations.map(toPath)))
+			return new ChokidarWatcherWrapper(
+				chokidar.watch(locations.map(toPath), { usePolling: process.platform === 'win32' }),
+			)
 		},
 		writeFile(location, data, options) {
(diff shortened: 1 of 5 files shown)


packages/core/src/common/externals/NodeJsExternals.ts
```diff
@@ -118,7 +118,9 @@ export const NodeJsExternals: Externals = {
 			return fsp.unlink(toFsPathLike(location))
 		},
 		watch(locations) {
-			return new ChokidarWatcherWrapper(chokidar.watch(locations.map(toPath)))
+			return new ChokidarWatcherWrapper(
+				chokidar.watch(locations.map(toPath), { usePolling: process.platform === 'win32' }),
```
misode, line 122: `process` is not available in the browser, so a new external method needs to be added to test for the platform

misode (changes requested):

misode:
I tried this PR on the Gamemode 4 project and looking at the logs, after 10 minutes of waiting the file watcher still wasn't ready. There were 32K files in my workspace, so that's not the average project, but if we add this it should definitely be opt-in

packages/core/src/common/externals/NodeJsExternals.ts
TheAfroOfDoom: we use it above in `showFile` -- is that method just never called in the browser?

TheAfroOfDoom: we already have a separate `BrowserExternals` file which i believe handles this stuff. i think using `process.platform` is fine because it's exclusive to this `NodeJsExternals` file

https://github.com/SpyglassMC/Spyglass/blob/cfa732191f3896cdf8b65cca1a3351ad76e16680/packages/core/src/common/externals/BrowserExternals.ts#L151-L153

TheAfroOfDoom: i'm not sure we've verified if the file-lock issue is even a thing in browsers

misode: Oh you are totally right 🤦

Issue #1496, opened 2024-07-08
Error when attempting to exclude default item components using !


thesorcell:
![spyglassreport](https://github.com/SpyglassMC/Spyglass/assets/93102875/aac7d0d6-5a0c-477a-87fc-473aeb069aa0)

Spyglass does not yet support the exclusion of default item components. This functionality was added in 24w18a, and is functional in the 1.21 release.

NateKootstra:
I am unsure as to how I would go about testing this, however I believe that replacing the declaration of LegalResourceLocationCharacters in **packages/core/src/parser/resourceLocation.ts** with the following should solve this issue:

export const LegalResourceLocationCharacters = new Set([
	'a',
	'b',
	'c',
	'd',
	'e',
	'f',
	'g',
	'h',
	'i',
	'j',
	'k',
	'l',
	'm',
	'n',
	'o',
	'p',
	'q',
	'r',
	's',
	't',
	'u',
	'v',
	'w',
	'x',
	'y',
	'z',
	'0',
	'1',
	'2',
	'3',
	'4',
	'5',
	'6',
	'7',
	'8',
	'9',
	'_',
	'-',
	'.',
	'!',
])

NateKootstra:
I am unsure as to how one would go about testing this, however I believe that replacing the declaration of LegalResourceLocationCharacters in **packages/core/src/parser/resourceLocation.ts** with the following should solve this issue:

`export const LegalResourceLocationCharacters = new Set([
'a',
'b',
'c',
'd',
'e',
'f',
'g',
'h',
'i',
'j',
'k',
'l',
'm',
'n',
'o',
'p',
'q',
'r',
's',
't',
'u',
'v',
'w',
'x',
'y',
'z',
'0',
'1',
'2',
'3',
'4',
'5',
'6',
'7',
'8',
'9',
'_',
'-',
'.',
'!',
])`

If anyone knows how to go about testing this change, please let me know. While I'm fairly confident that this should fix the issue, I'm not sure what the ramifications of this solution would be, and there is likely a better one. I would love to test it further.

NateKootstra:
After looking into the issue a bit more, I think that this solution should work. If anyone would like to verify whether or not this would work it'd be greatly appreciated. Unless someone points out a flaw or proposes a better solution I plan on opening a pull request.

PR #1539, opened 2024-07-20
Allow for exclamation marks in resource locations.

NateKootstra:
Replacing the initialization of LegalResourceLocationCharacters in `packages/core/src/parser/resourceLocation.ts` with the following should solve issue #1496:

`export const LegalResourceLocationCharacters = new Set([
	'a',
	'b',
	'c',
	'd',
	'e',
	'f',
	'g',
	'h',
	'i',
	'j',
	'k',
	'l',
	'm',
	'n',
	'o',
	'p',
	'q',
	'r',
	's',
	't',
	'u',
	'v',
	'w',
	'x',
	'y',
	'z',
	'0',
	'1',
	'2',
	'3',
	'4',
	'5',
	'6',
	'7',
	'8',
	'9',
	'_',
	'-',
	'.',
	'!',
])`

This should allow for exclamation marks to be used to remove components from items without Spyglass listing it as an error, which would be in line with how it works in game.

 In theory this will allow examples such as "!minecraft:food" to function, although "!food" would likely still result in a warning due to Spyglass interpreting it as "minecraft:!food". If anyone more experienced with Spyglass knows to fix that it would definitely improve the solution. I'll spend some more time trying to figure it out myself.

### DIFF
--- a/packages/core/src/common/util.ts
+++ b/packages/core/src/common/util.ts
@@ -129,7 +129,7 @@ export namespace ResourceLocation {
 	export function lengthen(value: string): FullResourceLocation {
 		switch (value.indexOf(NamespacePathSep)) {
 			case -1:
-				return `!${DefaultNamespace}${NamespacePathSep}${value}`
+				return `${DefaultNamespace}${NamespacePathSep}${value}`
 			case 0:
 				return `${DefaultNamespace}${value}` as unknown as FullResourceLocation
 			default:
--- a/packages/core/src/common/util.ts
+++ b/packages/core/src/common/util.ts
@@ -117,7 +117,7 @@ export namespace ResourceLocation {
 	 */
 	export const TagPrefix = '#'
 	/**
-	 * The seperator of namespace and path.
+	 * The prefix for removing a component.
 	 */
 	export const NamespacePathSep = ':'
 	/**
--- a/packages/core/src/common/util.ts
+++ b/packages/core/src/common/util.ts
@@ -117,7 +117,7 @@ export namespace ResourceLocation {
 	 */
 	export const TagPrefix = '#'
 	/**
-	 * The prefix for removing a component.
+	 * The seperator of namespace and path.
 	 */
 	export const NamespacePathSep = ':'
 	/**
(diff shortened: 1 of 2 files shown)


jacobsjo:
This is almost certainly the wrong place to fix this.

There are other places that use resource locations. These places do not allow exclamation marks. The exclamation mark is part of the syntax to remove components from items, its not part of the resource location itself.

NateKootstra:
After looking into it further, I think that adding `export const RemovalPrefix = '!'` to packages/core/src/common/util.ts and then making this modification to lengthen() in that same file should make "!food" work:

```
export function lengthen(value: string): FullResourceLocation {
	switch (value.indexOf(NamespacePathSep)) {
		case -1:
			return value.startsWith(RemovalPrefix) ? `!${DefaultNamespace}${NamespacePathSep}${value.replace(RemovalPrefix, '')}` : `!${DefaultNamespace}${NamespacePathSep}${value}`
		case 0:
			return (value.startsWith(RemovalPrefix) ? `!${DefaultNamespace}${value.replace(RemovalPrefix, '')}` : `${DefaultNamespace}${value}`) as unknown as FullResourceLocation
		default:
			return value as FullResourceLocation
	}
}
```

NateKootstra:
That's fair. Do you think that modifying `packages/core/src/common/util.ts` with something like this could work?

```
export function lengthen(value: string): FullResourceLocation {
	switch (value.indexOf(NamespacePathSep)) {
		case -1:
			return value.startsWith('!') ? `!${DefaultNamespace}${NamespacePathSep}${value.replace('!', '')}` : `!${DefaultNamespace}${NamespacePathSep}${value}`
		case 0:
			return (value.startsWith('!') ? `!${DefaultNamespace}${value.replace('!', '')}` : `${DefaultNamespace}${value}`) as unknown as FullResourceLocation
		default:
			return value as FullResourceLocation
	}
}
```

If not, what would the proper level to fix this at be?

NateKootstra:
Obviously after that we still need to allow for the exclamation mark in the FullResourceLocation.

NeunEinser:
This should be fixed in the item argument type specifically.

You can take a look at https://github.com/SpyglassMC/Spyglass/blob/main/packages/mcdoc/src/runtime/attribute/builtin.ts which includes a parser for resource locations in JSON / SNBT, which already works for !-prefixed item components in e.g. the set_components loot function.

Simply allowing an exclamation mark inside of resource locations is an inadequate solution for following reasons:
- item components are a registry, and you need to avoid the "undeclared symbol" warning when using a prefixed resource location
- auto complete should correctly suggest prefixed resource locations.

Something similar may also be implemented for the `type` selector argument, though I am not very familiar with that part of the code.

NateKootstra:
Thanks a ton for the help! I'll look into that when I get home.

NateKootstra:
Yeah to be honest I'm incredibly lost when it comes to navigating through this project. I'll give it another shot tomorrow. Where Spyglass actually start analyzing mcfunction components?

PR #1598, opened 2024-09-14
🐛 Remove worldgen-only heightmaps from autocomplete

### LINKED ISSUE #1590: Extra heightmap values in `positioned over` that Minecraft doesn't recognize
Spyglass defines `world_surface_wg` and `ocean_floor_wg` as available heightmaps in `execute positioned over`. These are not available in the actual sub-command (due to only existing during world generation).

image
image

Nico314159:
Fixes #1590


MulverineX (approved):
afaict vanilla-mcdoc takes care of the _wg stuff, so this should be good 👍

PR #1599, opened 2024-09-17
🐛 Use custom file tree walker instead of globby

### LINKED ISSUE #1572: Cannot find `pack.mcmeta` when projectRoot is "complex"
Example project root that doesn't work: `file:///c%3A/Users/foo/AppData/Roaming/.minecraft/saves/%C2%A7a%C2%A7lVersion%201.21%20%C2%A74%C2%A7l%5B4.5%5D%C2%A77/datapacks/bar/`, which after percent decoding is `file:///c:/Users/foo/AppData/Roaming/.minecraft/saves/§a§lVersion 1.21 §4§l[4.5]§7/datapacks/bar/`

Maybe all those percent encoding messed something up?

SPGoding:
Fix #1572.

### DIFF
--- a/packages/java-edition/src/index.ts
+++ b/packages/java-edition/src/index.ts
@@ -44,7 +44,7 @@ export const initialize: core.ProjectInitializer = async (ctx) => {
 		const searched = new Set<string>()
 		for (let depth = 0; depth <= 2; depth += 1) {
 			for (const projectRoot of projectRoots) {
-				const files = await externals.fs.getAllFiles(projectRoot, depth + 1)
+				const files = await core.fileUtil.getAllFiles(externals, projectRoot, depth + 1)
 				for (const uri of files.filter(uri => uri.endsWith('/pack.mcmeta'))) {
 					if (searched.has(uri)) {
 						continue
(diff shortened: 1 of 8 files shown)


packages/java-edition/src/index.ts
```diff
@@ -44,7 +44,7 @@ export const initialize: core.ProjectInitializer = async (ctx) => {
 		const searched = new Set<string>()
 		for (let depth = 0; depth <= 2; depth += 1) {
 			for (const projectRoot of projectRoots) {
-				const files = await externals.fs.getAllFiles(projectRoot, depth + 1)
+				const files = await core.fileUtil.getAllFiles(externals, projectRoot, depth + 1)
```
misode, line 47: The `getAllFiles` function here is called multiple times, each time increasing its depth, therefore repeating the directory walk. It may be possible to optimize this in the future.

misode (approved):
Testing this, I can confirm it fixes the "cannot find pack.mcmeta" issue.

However it does not fix the more general issue where symbols are not found. After some debugging I realized this is a completely separate problem, so I'll create a new issue for it.

packages/java-edition/src/index.ts
SPGoding: I think this wouldn't be necessary if we make `getAllFiles()` breadth-first instead of depth-first, so that the earliest `pack.mcmeta` file would be the shallowest, and we can limit depth to `2` directly.

misode: The point would also be to prevent checking depth 2 if there is a valid pack.mcmeta at depth 1

SPGoding: Hmm. It could be made into a breadth-first generator?

misode: But imo that isn't necessary to be done now. This will most likely already be more efficient than globby. So feel free to merge if you are fine with this.

SPGoding: Yeah. I also think if it were implemented it wouldn't necessarily be an optimization. We'd just be trading more memory consumption for less IO calls.

PR #1602, opened 2024-09-17
🐛 Disable chokidar globbing

### LINKED ISSUE #1601: Chokidar file watcher fails when projectRoot is "complex"
- Relates to  #1572 

When the project root is something like `file:///c%3A/Users/User/Documents/sandbox/Foo%20Bar%20%5Bv1.310%5D/`, this is converted to `c:\\Users\\User\\Documents\\sandbox\\Foo Bar [v1.310]\\` before being passed to Chokidar.

misode:
It turned out to be a really simple fix (excluding the work in #1599)

Fixes #1601


SPGoding (approved):
Nice lol.

PR #1607, opened 2024-09-27
⚡️ provide `exclude` patterns to file watcher args

TheAfroOfDoom:
right now we filter files for Spyglass to care about after we've already told chokidar to watch all our root directories.

chokidar ends up watching every file in the directory and is very slow, even if we later ignore certain files based on `config.env.exclude`.

not only that, but i couldn't get exclude patterns to work that you'd expect to work. for instance:

---

#### `datapacks/omega-flowey/data/*/test` wouldn't match the corresponding `test` directory relative to any of the project roots (mostly fixed ✅)

so i'd get lots of spyglass errors due to `packtest`'s non-vanilla commands just on project open

| before | after |
|-|-|
| image |image |

Spyglass still shows errors if you open the file, but this is progress.

we could fix ^ by hooking the exclude config into `connection.onHover` (etc.) somewhere, but that's out of scope here

---

#### explicitly ignored datapack directories would still get parsed and show up in references/definitions (also mostly fixed ✅)

i have a `build` directory that is `.gitignored`, and this would still show up when i open a function's references.

on project load, these files are no longer bound to the function references, but they unfortunately still are bound again if you open the file. still an improvement though, and we can probably remedy further with similar methods to the `packtest` related stuff above

| before, project load -> parses the explicitly `.gitignore`'d directory |
|-|
| image |

| after, project load -> doesnt parse (still parses if opened though) |
|-|
| image |

---

#### `resourcepack/pack.mcmeta` (from project root(s)) wouldn't match a RP pack.mcmeta; i'd still get pack version errors ⚠️ 

note that while the watcher is no longer watching this file if i add it to my exclude list, Spyglass still finds it because `findPackMcmeta` doesn't use the exclude config yet.

maybe less relevant though bc i know Misode's been working on improving that specific bug anyway

---

### major performance gain

regardless of the above issues being partially remedied, the biggest gain we get is from chokidar not recursively adding every file under every project root to the watched files list. with a properly formatted `ignored` list, chokidar stops early at a directory that matches an ignore pattern.

profile comparisons:

#### `#ready` time

| before, gitignore specified but broken parsing | after |
|-|-|
|image|image|

#### watch file count

| before, gitignore specified but broken parsing | after |
|-|-|
|image|image|

### DIFF
--- a/packages/core/src/common/externals/index.ts
+++ b/packages/core/src/common/externals/index.ts
@@ -72,7 +72,10 @@ export interface ExternalFileSystem {
 	showFile(path: FsLocation): Promise<void>
 	stat(location: FsLocation): Promise<{ isDirectory(): boolean; isFile(): boolean }>
 	unlink(location: FsLocation): Promise<void>
-	watch(locations: FsLocation[], options: { usePolling?: boolean }): FsWatcher
+	watch(
+		locations: FsLocation[],
+		options: { ignored?: string[]; usePolling?: boolean },
+	): FsWatcher
 	/**
 	 * @param options `mode` - File mode bit mask (e.g. `0o775`).
 	 */
--- a/packages/core/src/common/externals/index.ts
+++ b/packages/core/src/common/externals/index.ts
@@ -1,3 +1,4 @@
+import type { WatchOptions } from 'chokidar'
 import type { Uri } from '../util.js'
 import type { ExternalDownloader } from './downloader.js'
 
@@ -74,7 +75,7 @@ export interface ExternalFileSystem {
 	unlink(location: FsLocation): Promise<void>
 	watch(
 		locations: FsLocation[],
-		options: { ignored?: string[]; usePolling?: boolean },
+		options: WatchOptions,
 	): FsWatcher
 	/**
 	 * @param options `mode` - File mode bit mask (e.g. `0o775`).
--- a/packages/core/src/service/Project.ts
+++ b/packages/core/src/service/Project.ts
@@ -185,6 +185,7 @@ export class Project implements ExternalEventEmitter {
 	config!: Config
 	ignore: Ignore = ignore()
 	readonly downloader: Downloader
+	readonly #excludePaths: string[] = []
 	readonly externals: Externals
 	readonly fs: FileService
 	readonly isDebugging: boolean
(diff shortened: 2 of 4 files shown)


packages/core/src/common/externals/index.ts
```diff
@@ -72,7 +73,10 @@ export interface ExternalFileSystem {
 	showFile(path: FsLocation): Promise<void>
 	stat(location: FsLocation): Promise<{ isDirectory(): boolean; isFile(): boolean }>
 	unlink(location: FsLocation): Promise<void>
-	watch(locations: FsLocation[], options: { usePolling?: boolean }): FsWatcher
+	watch(
+		locations: FsLocation[],
+		options: WatchOptions,
```
misode, line 78: This shouldn't directly use chokidar's `WatchOptions`, this is an interface that needs to be shared between the nodejs and browser externals

packages/core/src/service/Project.ts
```diff
@@ -395,16 +398,30 @@ export class Project implements ExternalEventEmitter {
...
+						paths.push(...gitignorePaths)
 					}
 				} else {
-					this.ignore.add(pattern)
+					paths.push(pattern)
 				}
 			}
+			// Chokidar never matches paths with a trailing (back)slash, so fix any paths
+			// that may be specified as such
+			this.#excludePaths.push(...paths.map((path) => path.replace(/(\\|\/)$/, '')))
```
misode, line 424: Since this is a chokidar specific requirement, this should be moved to the `NodeJsExternals` `watch` handler.

packages/core/src/service/Project.ts
```diff
@@ -185,6 +187,7 @@ export class Project implements ExternalEventEmitter {
 	config!: Config
 	ignore: Ignore = ignore()
 	readonly downloader: Downloader
+	readonly #excludePaths: string[] = []
```
misode, line 190: This is weirdly named, it should be `excludePatterns`

packages/core/src/service/Project.ts
```diff
@@ -514,9 +531,36 @@ export class Project implements ExternalEventEmitter {
...
 					return
 				}
+
+				const ignored = []
+				const absoluteExcludePaths = this.#excludePaths
+					.filter(isAbsolute)
+					.map((path) => resolvePath(path))
+				const matchesAbsolutePath = (path: string): boolean => {
+					return absoluteExcludePaths.some((exclude) => {
+						if (process.platform === 'win32') {
```
misode, line 541: This will break in the browser

misode (changes requested):

TheAfroOfDoom:
TODO should maybe also bump cache version

misode:
No longer relevant after #1695 was fixed. I did experiment briefly with passing the patterns to Chokidar, but passing a function made performance worse (because `shouldExclude` is not very cheap), and passing the raw patterns to be used by anymatch didn't work. Maybe this should be attempted again, but since the plan is to remove Chokidar (#1610), I don't think we should prioritize that.

PR #1702, opened 2025-01-01
📝 Update README after the rewrite was finished

Loweredgames:


README.md
```diff
@@ -2,11 +2,6 @@
 
 This project is undergoing a complete rewrite. We're almost there.
```
NeunEinser, line 4: The whole Under construction and "this project is undergoing a complete rewrite, we're almost there" can go honestly. We are already there for a good while now :D

misode (approved):
Thank you!

PR #1708, opened 2025-01-04
📌 Fix package-lock.json

SPGoding:


misode:
As discussed on Discord, it turns out maybe this isn't necessary, as #1701 upgraded the `lockfileVersion` from 2 to 3 and it likely just optimized the package lock.

PR #1716, opened 2025-01-09
📝 Add json schema for spyglass.json config

legopitstop:


misode (approved):
Ideally we would have something that generates this from the existing type definitions, but this is a good way to improve the situation right now. Thanks for contributing!

SPGoding (approved):

PR #1746, opened 2025-02-07
✨ Integrate Spyglass web API

### LINKED ISSUE #1480: Proxy for api.github.com
Would hopefully prevent #1466 and any potential rate limit (like [this complaint in the Spyglass #general Discord channel](https://discord.com/channels/666020457568403505/666020458453270535/1257075540590854334) from happening.

### LINKED ISSUE #1734: Unexpected error when multiple GitHub webhooks for the same repository are delivered around the same time

SPGoding:
* Changed to download resources from `api.spyglassmc.com` instead of from GitHub.
* Replaced the old `Downloader` with something using the `Cache` API and `fetch()`. Implemented a very simple shim of the `Cache` API in `NodeJsExternals.ts`.

Fixes #1734 and resolves #1480.


packages/language-server/src/server.ts
```diff
@@ -81,7 +81,7 @@ connection.onInitialize(async (params) => {
 				defaultConfig: core.ConfigService.merge(core.VanillaConfig, {
 					env: { gameVersion: initializationOptions?.gameVersion },
 				}),
-				cacheRoot: fileUtil.ensureEndingSlash(url.pathToFileURL(cacheRoot).toString()),
+				cacheRoot,
 				externals,
```
misode, line 85: If `cacheRoot` is now given when creating `externals`, is it still necessary to store this separately in `Project`?

packages/java-edition/src/dependency/index.ts
```diff
@@ -83,179 +64,106 @@ export async function getMcmetaSummary(
...
- * 	- `startDepth`: The amount of level to skip when unzipping the tarball.
- * 	- `uri`: URI to the `.tar.gz` file.
  */
 export async function getVanillaDatapack(
-	downloader: core.Downloader,
+	externals: core.Externals,
+	logger: core.Logger,
 	version: string,
-	isLatest: boolean,
 ): Promise<core.Dependency> {
```
misode, line 108: `isLatest` can also be removed entirely from `McmetaVersion`

misode (changes requested):

packages/language-server/src/server.ts
SPGoding: The fact that we pass `cacheRoot` to `externals` is more of a side effect of how `NodeJsExternals` functions -- it needs a path to store the cached responses under. It won't exist if `BrowserExternals` is used, for example, so I'd still like to keep passing it to `Project` as well.

misode: Oh I guess it is still used to cache the symbol table and virtual uris

misode (changes requested):
Testing this branch locally, a lot (but not all) mcdoc symbols are borked.
image

Looking at the warning paths, maybe the `stripLevel` is incorrect now?

misode (commented):
I experienced a corrupt vanilla-data cache once during unstable internet, but I can't really reproduce this.

I also experienced an error once that said `versions.sort` is not a function (in `resolveConfiguredVersion`), this would indicate that `versions` was not underfined, but also not a list, which seems really strange. Writing this I realized this might've been an `{ "message": "..." }`, so maybe `fetcher.ts` is not handling 4xx and 5xx responses correctly

misode (approved):

packages/core/src/service/fetcher.ts
```diff
@@ -15,7 +15,9 @@ export async function fetchWithCache(
 		request.headers.set('If-None-Match', cachedEtag)
 	}
 	try {
-		const response = await web.fetch(request)
+		const response = await web.fetch(request, {
+			headers: { 'User-Agent': 'SpyglassMC (+https://spyglassmc.com)' },
```
misode, line 19: Perhaps we could include the vscode extension version number here? Or if that is too impractical, maybe the core package version?

SPGoding: I wonder how to get that number at runtime though, since depending on how the package is used/bundled the location of `package.json` (if it even exists) could anywhere. I guess we could store it as a constant somewhere and modify it in the `release` script?

packages/core/src/service/fetcher.ts
```diff
@@ -0,0 +1,50 @@
...
+			return cachedResponse
+		} else if (!response.ok) {
+			let message = response.statusText
+			try {
+				message = (await response.json()).message
+			} catch (e) {}
+			throw new TypeError(`${response.status} ${message}`)
+		} else {
+			try {
+				await cache.put(request, response)
```
misode, line 33: I noticed an error here when using this code in the browser. Over there it actually consumes the response body when calling `cache.put`, without cloning. So instead of putting `response.clone()` inside `HttpCache`, it should be put in this function.

misode (changes requested):

misode (approved):

PR #1751, opened 2025-02-10
Do not send documents for archive urls (Fixes #1745)

### LINKED ISSUE #1745: Language server reports diagnostics for source archives
Hey, I started working on a plugin for JetBrains that uses this LSP: https://github.com/JulianVennen/spyglass-mc-jetbrains

While working on that, I noticed numerous errors in my IntelliJ log that looked something like this:
```
2025-02-04 18:47:42,428 [  10441]   WARN - #c.i.p.l.a.LspServerDescriptor - Unexpected URI scheme: archive://vanilla-mcdoc.tar.gz/java/server/world/entity/mob/breedable/tamable.mcdoc
```

After enabling debug logging which outputs all LSP messages, the following appeared:
```
2025-02-04 18:47:42,428 [  10441]   FINE - #c.i.p.l.i.c.Lsp4jServerConnector - <-- SpyglassLspServerDescriptor@towers: {"jsonrpc":"2.0","method":"textDocument/publishDiagnostics","params":{"diagnostics":[],"uri":"archive://vanilla-mcdoc.tar.gz/java/server/world/entity/mob/fish.mcdoc"}}
```

Spyglass appears to be reporting diagnostics for files inside the vanilla-mcdoc.tar.gz and mcmeta.tar.gz. Obviously, files inside these are not part of the project I have opened in my IDE, so it makes sense that IntelliJ logs an error in that case.

JulianVennen:
Fixes #1745

### DIFF
--- a/packages/language-server/src/server.ts
+++ b/packages/language-server/src/server.ts
@@ -88,7 +88,7 @@ connection.onInitialize(async (params) => {
 			},
 		})
 		service.project.on('documentErrored', async ({ errors, uri, version }) => {
-			if (!uri.startsWith("file://")) {
+			if (!uri.startsWith('file://')) {
 				return
 			}
 
--- a/packages/language-server/src/server.ts
+++ b/packages/language-server/src/server.ts
@@ -88,6 +88,10 @@ connection.onInitialize(async (params) => {
 			},
 		})
 		service.project.on('documentErrored', async ({ errors, uri, version }) => {
+			if (!uri.startsWith("file://")) {
+				return
+			}
+
 			try {
 				await connection.sendDiagnostics({
 					diagnostics: toLS.diagnostics(errors),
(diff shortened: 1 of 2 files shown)


SPGoding (approved):

packages/language-server/src/server.ts
```diff
@@ -88,6 +88,10 @@ connection.onInitialize(async (params) => {
 			},
 		})
 		service.project.on('documentErrored', async ({ errors, uri, version }) => {
+			if (!uri.startsWith('file://')) {
```
misode, line 91: Is it a good idea to explicitly check for `file` rather than filter out `archive` explicitly? Are we sure this won't break stuff like remote ssh, live share, or others?

JulianVennen: Both the intellij built in lsp client and lsp4intellij check explicitly for file://. If there's a case where vscode accepts other urls I can check for archive:// instead

SPGoding: Afaik VS Code uses the `untitled:` URI scheme for new files that are not saved on disk

misode (approved):

SPGoding (approved):

PR #1755, opened 2025-02-15
Extend version attribute

Flemmli97:
Extends the attributes for versions adding
- `only`: To select a specific version
- `except`: To select a version range outside of the specified one. Syntax `min-max` 

Currently you can mimic `only` with `since` and `until` but not `except`.

### DIFF
--- a/packages/java-edition/src/mcdocAttributes.ts
+++ b/packages/java-edition/src/mcdocAttributes.ts
@@ -33,6 +33,26 @@ export function registerMcdocAttributes(
 			return ReleaseVersion.cmp(release, config as ReleaseVersion) < 0
 		},
 	})
+	mcdoc.runtime.registerAttribute(meta, 'only', validator.string, {
+		filterElement: (config, ctx) => {
+			if (!config.startsWith('1.')) {
+				ctx.logger.warn(`Invalid mcdoc attribute for "until": ${config}`)
+				return true
+			}
+			return ReleaseVersion.cmp(release, config as ReleaseVersion) == 0
+		},
+	})
+	mcdoc.runtime.registerAttribute(meta, 'between', validator.string, {
+		filterElement: (config, ctx) => {
+			const range = config.split("-")
+			if (range.length != 2 || !range[0].startsWith('1.')|| !range[1].startsWith('1.')) {
+				ctx.logger.warn(`Invalid mcdoc attribute for "until": ${config}`)
+				return true
+			}
+			return ReleaseVersion.cmp(release, range[0] as ReleaseVersion) >= 0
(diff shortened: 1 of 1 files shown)


misode:
I'm currently not sure whether I like these attributes or not. I would have to think about it a bit more. They wouldn't be particularly used a lot in vanilla-mcdoc, so I wonder if you have a use case for these?

Flemmli97:
it would be very niche i agree. and would most likely only apply to modded generators. i currently have a mod that got a bit of changes in 1.18 that i plan to also port to 1.21 (but not the 1.20 version most likely)

packages/java-edition/src/mcdocAttributes.ts
```diff
@@ -33,6 +33,26 @@ export function registerMcdocAttributes(
 			return ReleaseVersion.cmp(release, config as ReleaseVersion) < 0
 		},
 	})
+	mcdoc.runtime.registerAttribute(meta, 'only', validator.string, {
+		filterElement: (config, ctx) => {
+			if (!config.startsWith('1.')) {
+				ctx.logger.warn(`Invalid mcdoc attribute for "until": ${config}`)
```
NeunEinser, line 39: This should say "only" instead of "until".

packages/java-edition/src/mcdocAttributes.ts
```diff
@@ -33,6 +33,26 @@ export function registerMcdocAttributes(
...
+				return true
+			}
+			return ReleaseVersion.cmp(release, config as ReleaseVersion) == 0
+		},
+	})
+	mcdoc.runtime.registerAttribute(meta, 'except', validator.string, {
+		filterElement: (config, ctx) => {
+			const range = config.split("-")
+			if (range.length != 2 || !range[0].startsWith('1.')|| !range[1].startsWith('1.')) {
```
NeunEinser, line 48: Other than the linting issue the linter points out, there is a missing space before one of the `||`

packages/java-edition/src/mcdocAttributes.ts
```diff
@@ -33,6 +33,26 @@ export function registerMcdocAttributes(
...
+				ctx.logger.warn(`Invalid mcdoc attribute for "until": ${config}`)
```
NeunEinser, line 49: This should say "except" instead of "until".

NeunEinser (commented):
I'm also not too sure about these attributes. Maybe something that cleanly lets you specify multiple ranges instead?

This would break down again, if you had multiple version ranges that should be excluded.

Flemmli97:
what about something like:
`min-max,min2-max2 ...`?
and maybe also `min..` or `..max`? though this would look ugly since versions already have dots in them.

this approach would also make `since` and `until` redundant so we could remove that (which would be breaking change though)

PR #1758, opened 2025-02-16
🐛 Ignore non-Minecraft JSON files

SPGoding:
(This PR could have each commit individually merged instead of a squash merge since I separated units of change into four individually revertable commits. Well after a `git rebase --autosquash -i 3ea96b` that is)

This PR adds an additional `uriPredicate` option to `LanguageOptions` during language registration. The `json` language definition has been updated to include a `uriPredicate` to ensure the JSON file actually belongs to a data pack or a resource pack. This check is more accurate than the `DocumentFilter` implemented client side in the VS Code extension as it takes the roots and the available file categories into consideration. Additions have also been made to `Project.shouldExclude()` to make sure unsupported files never go into the document life cycle.

This PR also fixes two bugs I noticed while making the changes.

### DIFF
--- a/packages/core/src/service/Project.ts
+++ b/packages/core/src/service/Project.ts
@@ -872,8 +872,8 @@ export class Project implements ExternalEventEmitter {
 		content: string,
 	): Promise<void> {
 		uri = this.normalizeUri(uri)
-		if (!fileUtil.isFileUri(uri)) {
-			return // We only accept `file:` scheme for client-managed URIs.
+		if (uri.startsWith(ArchiveUriSupporter.Protocol)) {
+			return // We do not accept `archive:` scheme for client-managed URIs.
 		}
 		if (this.shouldExclude(uri)) {
 			return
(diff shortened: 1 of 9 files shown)


packages/core/src/service/Project.ts
```diff
@@ -287,10 +288,7 @@ export class Project implements ExternalEventEmitter {
 	 * are not loaded into the memory.
 	 */
 	getTrackedFiles(): string[] {
-		const extensions: string[] = this.meta.getSupportedFileExtensions()
-		this.logger.info(`[Project#getTrackedFiles] Supported file extensions: ${extensions}`)
 		const supportedFiles = [...this.#dependencyFiles ?? [], ...this.#watchedFiles]
-			.filter((file) => extensions.includes(fileUtil.extname(file) ?? ''))
 		this.logger.info(
 			`[Project#getTrackedFiles] Listed ${supportedFiles.length} supported files`,
```
SPGoding, line 293: The extension check here has been removed as all files in `#dependencyFiles` and `#watchedFiles` should now all be of supported languages already.

packages/java-edition/src/binder/index.ts
```diff
@@ -226,41 +274,7 @@ export function dissectUri(uri: string, ctx: UriBinderContext) {
...
-					continue
-				}
-				identifier = identifier.slice(0, -res.ext.length)
-				if (res.identifier && identifier !== res.identifier) {
-					continue
-				}
-				candidateResources.push([res, identifier])
-			}
-		}
+		const candidateResources = getCandidateResourcesForRel(rel)
```
misode, line 277: The logic between `const parts = res.split('/')` and here is also duplicated in `getCandidateResourcesForRel`

misode (changes requested):

packages/java-edition/src/binder/index.ts
misode: Hmm I see. Though it already returns a list of two things: `function getCandidateResourcesForRel(rel: string): [res: Resource, identifier: string][]`, so maybe it wouldn't be that bad to have it also return the namespace in that tuple (or change to an object)?

misode (approved):

PR #1760, opened 2025-02-16
Add match regex attribute

Flemmli97:
Adds an attribute `match_regex` that allows for regex validation for string fields.
Regex config is required but error is not. If error is not provided will use a default error message instead.
`#[match_regex=(regex=foo, error="error")]`

### DIFF
--- a/packages/mcdoc/src/runtime/attribute/builtin.ts
+++ b/packages/mcdoc/src/runtime/attribute/builtin.ts
@@ -13,6 +13,11 @@ interface IdConfig {
 	exclude?: string[]
 }
 
+interface RegexConfig {
+	regex: string
+	error?: string
+}
+
 const idValidator = validator.alternatives<IdConfig>(
 	validator.map(validator.string, v => ({ registry: v })),
 	validator.tree({
@@ -267,6 +272,38 @@ export function registerBuiltinAttributes(meta: core.MetaRegistry) {
 			return (node, ctx) => {
 				try {
 					RegExp(pattern)
+				} catch (e) {
+					const message = e instanceof Error ? e.message : `${e}`
+					const error = message
+						.replace(/^Invalid regular expression: /, '')
+						.replace(/^\/.+\/: /, '')
+					ctx.err.report(localize('invalid-regex-pattern', error), node, 2)
+				}
+			}
+		},
+	})
+	registerAttribute(meta, 'match_regex', () => undefined, {
+		checker: (config, typeDef, _) => {
+			if (typeDef.kind !== 'literal' || typeDef.value.kind !== 'string') {
+				return undefined
+			}
+			const conf = config as RegexConfig
+			const pattern = conf.regex
+			const value = typeDef.value.value
+			return (node, ctx) => {
+				try {
+					const regex = RegExp(pattern)
+					if (!regex.test(value)) {
+						if (conf.error) {
+							ctx.err.report(conf.error, node, 2)
+						} else {
+							ctx.err.report(
+								localize('mismatching-regex-pattern', typeDef.value.value),
+								node,
+								2,
+							)
+						}
+					}
 				} catch (e) {
 					const message = e instanceof Error ? e.message : `${e}`
 					const error = message
--- a/packages/locales/src/locales/en.json
+++ b/packages/locales/src/locales/en.json
@@ -48,6 +48,7 @@
   "integer.between": "an integer between %0% and %1%",
   "invalid-key-combination": "Invalid combination of keys %0%",
   "invalid-regex-pattern": "Invalid regex pattern: %0%",
+  "mismatching-regex-pattern": "Value does not match regex: %0%",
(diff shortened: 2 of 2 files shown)


NeunEinser (approved):
Lgtm, only thing is, I am not sure if the `error` message would actually work like this if we are thinking of translatable mcdoc. But I guess it's not that relevant for vanilla-mcdoc anyways.

misode:
I think it's acceptible that the custom error message is not translatable

packages/mcdoc/src/runtime/attribute/builtin.ts
```diff
@@ -30,6 +30,19 @@ const idValidator = validator.alternatives<IdConfig>(
...
+interface RegexConfig {
+	regex: string
+	error?: string
+}
+
+const regexMatchValidator = validator.alternatives<RegexConfig>(
+	validator.map(validator.string, v => ({ regex: v })),
+	validator.tree({
+		regex: validator.string,
+		error: validator.optional(validator.string),
```
misode, line 42: I would call these fields `pattern` and `message` personally.

packages/mcdoc/src/runtime/attribute/builtin.ts
```diff
@@ -277,4 +290,35 @@ export function registerBuiltinAttributes(meta: core.MetaRegistry) {
...
+				try {
+					const regex = RegExp(pattern)
+					if (!regex.test(value)) {
+						if (config.error) {
+							ctx.err.report(config.error, node, 2)
+						} else {
+							ctx.err.report(
+								localize('mismatching-regex-pattern', typeDef.value.value),
+								node,
+								2,
```
misode, line 310: Use `core.ErrorSeverity.Warning` instead of `2` here. I know this file already used `2` for this but that was probably a mistake.

packages/mcdoc/src/runtime/attribute/builtin.ts
```diff
@@ -272,7 +285,46 @@ export function registerBuiltinAttributes(meta: core.MetaRegistry) {
...
+				return undefined
+			}
+			const pattern = config.pattern
+			const value = typeDef.value.value
+			return (node, ctx) => {
+				try {
+					const regex = RegExp(pattern)
+					if (!regex.test(value)) {
+						if (config.message) {
+							ctx.err.report(config.message, node, 2)
```
SPGoding, line 309:
```suggestion
							ctx.err.report(config.message, node, core.ErrorSeverity.Warning)
```

You missed this :p

packages/mcdoc/src/runtime/attribute/builtin.ts
```diff
@@ -272,7 +285,46 @@ export function registerBuiltinAttributes(meta: core.MetaRegistry) {
...
+							)
+						}
+					}
+				} catch (e) {
+					const message = e instanceof Error ? e.message : `${e}`
+					const error = message
+						.replace(/^Invalid regular expression: /, '')
+						.replace(/^\/.+\/: /, '')
+					ctx.err.report(
+						localize('invalid-regex-pattern', error),
```
misode, line 324: I don't think this error should be reported. If the mcdoc type definitions contain a malformed regex pattern, the end-user shouldn't be bothered by that. Similar to some other attribute handling, this should just be ignored. At most it could be logged.

packages/mcdoc/src/runtime/attribute/builtin.ts
```diff
@@ -30,6 +30,19 @@ const idValidator = validator.alternatives<IdConfig>(
 	() => ({}),
 )
 
+interface RegexConfig {
+	pattern: string
+	message?: string
+}
+
+const regexMatchValidator = validator.alternatives<RegexConfig>(
```
misode, line 38: Should probably be named `matchRegexValidator` and `MatchRegexConfig` for consistency

packages/mcdoc/src/runtime/attribute/builtin.ts
```diff
@@ -272,7 +285,46 @@ export function registerBuiltinAttributes(meta: core.MetaRegistry) {
...
+			const value = typeDef.value.value
+			return (node, ctx) => {
+				try {
+					const regex = RegExp(pattern)
+					if (!regex.test(value)) {
+						if (config.message) {
+							ctx.err.report(config.message, node, core.ErrorSeverity.Warning)
+						} else {
+							ctx.err.report(
+								localize('mismatching-regex-pattern', typeDef.value.value),
```
misode, line 312: You can use `const message = config.message ?? localize('mismatching-regex-pattern', typeDef.value.value)` here to simplify this to a single `ctx.err.report` call

misode (changes requested):

packages/mcdoc/src/runtime/attribute/builtin.ts
misode: `ctx.logger` messages shouldn't be translated, and also include some info that this is related to the `match_regex` mcdoc info. See the logs for `since` and `until` as reference.

misode (approved):

SPGoding (approved):

PR #1762, opened 2025-02-17
🐛 Fix hash of binary files in cache service (attempt 2)

misode:
Same as #1707

But that PR created some issues where too many file handles were open (or something like that). I still think most of this PR are good changes, but might need some reworking


SPGoding (approved):
I'm still not exactly sure how the "too many files open" error occurred since from what I understand we're opening files in sequence, so there should never be multiple file handles open at the same time.

PR #1768, opened 2025-02-22
Extend version attribute

Flemmli97:
Extends the attributes for versions adding
- `only`: To select a specific version
- `except`: To select a version range outside of the specified one. Syntax `min-max` 

Currently you can mimic `only` with `since` and `until` but not `except`.

see #1755
new due to branch stuff on fork

### DIFF
--- a/packages/java-edition/src/mcdocAttributes.ts
+++ b/packages/java-edition/src/mcdocAttributes.ts
@@ -33,6 +33,26 @@ export function registerMcdocAttributes(
 			return ReleaseVersion.cmp(release, config as ReleaseVersion) < 0
 		},
 	})
+	mcdoc.runtime.registerAttribute(meta, 'only', validator.string, {
+		filterElement: (config, ctx) => {
+			if (!config.startsWith('1.')) {
+				ctx.logger.warn(`Invalid mcdoc attribute for "until": ${config}`)
+				return true
+			}
+			return ReleaseVersion.cmp(release, config as ReleaseVersion) == 0
+		},
+	})
+	mcdoc.runtime.registerAttribute(meta, 'between', validator.string, {
+		filterElement: (config, ctx) => {
+			const range = config.split("-")
+			if (range.length != 2 || !range[0].startsWith('1.')|| !range[1].startsWith('1.')) {
+				ctx.logger.warn(`Invalid mcdoc attribute for "until": ${config}`)
+				return true
+			}
+			return ReleaseVersion.cmp(release, range[0] as ReleaseVersion) >= 0
+				&& ReleaseVersion.cmp(release, range[1] as ReleaseVersion) <= 0
+		},
+	})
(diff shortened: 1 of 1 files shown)


MulverineX (approved):

misode:
I'm going to oppose merging these attributes. I don't think they're a good idea, mostly for the following reason:

Currently there are only two attributes someone needs to implement to filter mcdoc types based on versions. Imo `since` and `until` are the two basic building blocks to describe any version constraint, and it can only ever be written in a single way.

Both `only` and `except` rarely occur in practice, and if they do you can already achieve it by writing a tiny bit more code. We shouldn't add attributes for every small use-case. If we do, what's stopping us from imagining `between`, etc in a couple months? We should try to stabilize the set of attributes instead of carelessly expanding them.

Flemmli97:
what about multiple version ranges?

from the closed pr:

what about something like:
`(min-max,min2-max2 ...)`?
and maybe also `min..` or `..max`? though this would look ugly since versions already have dots in them.

this approach would also make `since` and `until` redundant so we could remove that (which would be breaking change though)

misode:
I don't think it's necessary to add more complexity. The current version attributes have been sufficient for vanilla-mcdoc.

For example if someone currently wants to make a mcdoc visualization tool and have a simple "added in <version>" annotation, it's simple to do that by checking a single attribute, If we start adding a bunch of attributes that all do the same but with different formats, or we create a complex syntax for specifying version ranges that would complicate the task.

MulverineX:
In defense, `except` can be safely ignored (with some penalty), and `only` is *very* easy to implement

Flemmli97:
wouldnt be something like 
`range="min.."` = `since`
`range="..max"` = `until` be easier to use?
`range="min-max"` would then also work.

and unless im wrong then something like this would also be possible then:
```
#[range="min-max"]
#[range="min2-max2"]
foo
```

misode:
The way it works right now is that `since` is inclusive, and `until` is exclusive. This means that you are always writing the version where the particular change was made. You can interpret the current attributes as "added in version ..." and "removed in version ..." respectively.

Flemmli97:
that could still be implemented with range

Flemmli97:
> wouldnt be something like `range="min.."` = `since` `range="..max"` = `until` be easier to use? `range="min-max"` would then also work.
> 
> and unless im wrong then something like this would also be possible then:
> 
> ```
> #[range="min-max"]
> #[range="min2-max2"]
> foo
> ```

> that could still be implemented with range

thoughs on that^ ? if thats better i could change this pr to match it

MulverineX:
that doesn't address misode's point about complexity

MulverineX:
On second thought, this doesn't really make a whole lot of sense. Maybe it would've made more sense when we were still developing the language, but now since/until are pretty set in stone.

PR #1809, opened 2025-05-13
✨ Use vanilla required_level added in 25w20a

jacobsjo:
+ Adjusts patches for old version to use the same format.


misode:
Looking at the code here, I'm not a fan that `getPatch` doesn't end up returning the final correct patch, and that the it needs to be patched again after being merged with the raw commands dump.

I feel like a better solution would be to change our default in the patch to `0` as well and explicitly patch the 61 nodes that have `required_level: 2` for backwards compatibility.

MulverineX:
This should probably either get closed or get marked as a Draft

PR #1815, opened 2025-05-14
✨ Implement macro option in command mcdoc attribute

misode:
Adds support for `#[command(macro="implicit")]`. I had to modify the macro AST nodes to support macro nodes without the leading `$`.


MulverineX (approved):

Calverin (approved):
Smart moving the error checking out of the macro parser, things look cleaner overall now 👍

Calverin:
Should probably update the test code results before merging though

PR #1819, opened 2025-05-24
👷 Separate language-server build from vscode-extension

axelkar:


axelkar:
Oh does the VS Code extension's server runner still work? I changed the format in the language-server esbuild config from cjs to esm

axelkar:
You may want to take this in as well, but we'll have it running in Nixpkgs regardless.
https://github.com/axelkar/Spyglass/commit/8b330a360b5e49b55d539c146a740b0b9a0d8fae

PR #1820, opened 2025-05-26
✨ Mcdoc Formatter

### LINKED ISSUE #1828: Implement mcdoc formatter/printer

Papierkorb2292:
This PR adds `Formatter`s for mcdoc nodes. At the moment this PR does not re-enable the `documentFormattingProvider`, but to try out the formatter it makes sense to temporarily enable it locally.

There are still some improvements that have to be made / special cases that have to be handled to ensure consistently good formatting.

Resolves #1828


Papierkorb2292:
Small update for the sake of completeness: The special cases and issues mentioned in the first comment have been addressed and the formatter is now registered to vscode as well (but only for the mcdoc language)

MulverineX (changes requested):
This PR should contain a "Format Workspace" command addition to the mcdoc CLI, I'm going to work on this

packages/core/src/processor/formatter/builtin.ts
```diff
@@ -38,11 +38,15 @@ export const boolean: Formatter<BooleanBaseNode> = (node) => {
 }
 
 export const comment: Formatter<CommentNode> = (node) => {
-	return '#' + node.comment
+	return node.prefix + node.comment
 }
 
 export const float: Formatter<FloatBaseNode> = (node) => {
-	return node.value.toString()
+	return node.value.toLocaleString('fullwide', {
```
misode, line 45: I don't think we should use `toLocaleString` here

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString
> Note: Most of the time, the formatting returned by toLocaleString() is consistent. However, the output may vary between implementations, even within the same locale — output variations are by design and allowed by the specification.

packages/language-server/src/server.ts
```diff
@@ -407,6 +411,12 @@ connection.onWorkspaceSymbol(({ query }) => {
 })
 
 connection.onDocumentFormatting(async ({ textDocument: { uri }, options }) => {
+	const errors = service.project.cacheService.errors[uri]
+	if (errors && errors.some(error => error.severity === core.ErrorSeverity.Error)) {
+		// Don't format if there are errors.
```
misode, line 416: Instead of checking the `cacheService`, it seems better to me to check the `parseErrors` on the document's AST file node. Checker and linter errors shouldn't prevent a file from being formatted, and I also don't know whether the cacheService errors always fully match the AST.

packages/mcdoc/test/formatter/suites/unionWithTrailingComment.mcdoc
```diff
@@ -0,0 +1,3 @@
+type MyType = (int | string)
+
+// This comment will be parsed as part of the union, but it shouldn't make the union multiline since it comes after the union.
```
misode, line 3: Missing EOL

packages/language-server/src/server.ts
```diff
@@ -155,6 +153,12 @@ connection.onInitialize(async (params) => {
 })
 
 connection.onInitialized(async () => {
+	if (capabilities.textDocument?.formatting?.dynamicRegistration) {
+		void connection.client.register(
+			ls.DocumentFormattingRequest.type,
+			{ documentSelector: [{ language: 'mcdoc' }] },
```
misode, line 159: What's the main purpose of doing this as a dynamic registration? Is it not possible to configure the `documentFormattingProvider` above and give it a document selector?

misode (changes requested):
I haven't looked at or tried out the formatter code itself, but here are my comments so far

packages/core/src/processor/formatter/builtin.ts
Papierkorb2292: Oh, alright, although I don't think there are any better alternatives, so we'll have to separately add the first decimal place

Papierkorb2292:
> I haven't looked at or tried out the formatter code itself, but here are my comments so far

Thanks for the feedback, the float formatter, checking for errors and the EOL for trailing comments should be fixed now

PR #1898, opened 2025-11-16
✅ Use native node:test runner to run tests

SPGoding:
* Updates all test code to use `erasableSyntaxOnly` TypeScript so they can be executed natively by Node.js v22.18.0+ without transpiling or flags.
  * Did the same for the development scripts and removed the `ts-node` dependency.
* No longer emits JavaScript code for the tests, thus removing all references to the `test-out/` directories
* Switches to use native `node:test` runner and its snapshot testing feature (available since v22.3.0).
  * Removed `mocha`, `snap-shot-core`, and `snap-shot-it` dependencies. As a bonus of this, `npm audit` finally gives "found 0 vulnerabilities".
  * Tried my best to keep snapshot formats the same. There are unfortunately still some minor differences.
* Updates npm scripts and VS Code launch config
  * We finally have reliable watch mode tests now! `npm run test:watch`
  * Updating snapshots is a lot more straightforward with `npm run test:update-snapshots`

At least for me the test also somehow got faster:

| mocha | node test runner |
| - | - |
| image | image |

(I have no idea why there are 51 more tests in node test runner than mocha)


.eslintrc.cjs
```diff
@@ -63,6 +87,10 @@ module.exports = {
 				"selector": 'ImportDeclaration > Literal[value=/\\bsrc\\b/]',
 				"message": "Import from the `lib` dir instead of the `src` dir."
 			},
+			{
+				"selector": 'CallExpression[callee.name=/describe|it/] > ObjectExpression > Property[key.name="only"][value.raw="true"]',
+				"message": "Use { only: true } for local testing only. Remember to remove it before merging to main."
+			},
```
SPGoding: The node test runner doesn't have something equivalent to mocha's `--forbid-only` argument for running on CI/CD so these new rules have been added in ESLint to catch them here instead.

packages/core/src/source/LanguageError.ts
```diff
@@ -52,7 +52,7 @@ export namespace LanguageError {
 	}
 }
 
-export const enum ErrorSeverity {
+export enum ErrorSeverity {
```
SPGoding: `const enum` is weird. Code don't play nice with it under `"verbatimModuleSyntax": true`.

packages/core/test/snapshot-setup.ts
```diff
@@ -0,0 +1,31 @@
...
+	if (packagesIndex === -1) {
+		throw new Error(`Could not find '/packages/' directory in testPath: ${testPath}`)
+	}
+
+	return path.join(
+		testPath.slice(0, packagesIndex),
+		'__snapshots__',
+		testPath.slice(packagesIndex + 1)
+			// Directory is called /test-out/ for historical reasons
+			.replace(new RegExp(`${path.sep}test${path.sep}`), `${path.sep}test-out${path.sep}`)
```
SPGoding: Goal is to keep the new snapshot files at the exact same locations as the old ones, so decided to keep the `test-out` part in the snapshot file paths.

packages/core/test/snapshot-setup.ts
```diff
@@ -0,0 +1,31 @@
...
+			.replace(/\.spec\.(ts|mts|cts)$/, '.spec.js'),
+	)
```
SPGoding: This function worked fine for me on Linux. Would be nice to have some Windows users test it.

packages/core/test/utils.ts
```diff
@@ -70,10 +56,10 @@ export function mockProjectData(data: Partial<ProjectData> = {}): ProjectData {
 }
 
 /**
- * @returns The string with `\t`, `\r`, `\n`, and `\\` replaced with non-special characters.
+ * @returns The string with `\t`, `\r`, and `\n` replaced with non-special characters.
  */
 export function showWhitespaceGlyph(string: string) {
-	return string.replace(/\t/g, '⮀').replace(/\r/g, '←').replace(/\n/g, '↓').replace(/\\/g, '⧵') // We replace normal back slashes with ⧵ (U+29f5) here, due to the snapshots being stupid and not escaping them before exporting.
+	return string.replace(/\t/g, '⮀').replace(/\r/g, '←').replace(/\n/g, '↓')
```
SPGoding: Finally no more weird slashes needed in the snapshot names

packages/core/test/snapshot-setup.ts
misode: Imo it would be acceptable to move them to the default location, now that it's been verified that the snapshots are correct.

.gitignore
```diff
@@ -1,19 +1,16 @@
...
+/.helix/languages.toml
 /.history
 /.vscode/settings.json
-/.mocharc.local.jsonc
 /.idea
 
 /packages/*/lib
 /packages/*/out
 /packages/*/dist
-/packages/*/test-out
```
misode: I think it's a good idea to keep all the old lines in here for a little bit, to avoid problems when switching branches. For example when I checked out this PR I had 266 pending changes from the `.nyc_output` and `test-out` folders

misode (changes requested):
I'm trying this on Windows 11 with node v25.1.0 and I'm getting an error for all snapshots
```
test at packages\mcfunction\test\parser\literal.spec.ts:14:3
✖ Parse 'advancement grant @s everything' (0.1191ms)
  SyntaxError: Invalid regular expression: /\test\/: \ at end of pattern
      at new RegExp (<anonymous>)
      at file:///C:/Users/<USER>/Documents/projects/spyglass/packages/core/test/snapshot-setup.ts:28:13
```

I messed with this a bit. If I removed the `test` to `test-out` replacing it still gave me a different error:
```
✖ __snapshots__\packages\nbt\test\parser\primitive.spec.js (0.025ms)
  ReferenceError: exports is not defined in ES module scope
  This file is being treated as an ES module because it has a '.js' file extension and 'C:\Users\<USER>\Documents\projects\spyglass\package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
```

This is possibly because I am using node 25. After changing the second rename call to `replace(/\.spec\.(ts|mts|cts)$/, '.spec.cjs')` (note the `cjs` extension), tests are working fine for me 🚀

.gitignore
SPGoding: Good call!

misode (approved):
Tests are working fine for me now.

Maybe slight nit: the `.only` eslint rule only works for me when I run `npm run lint`, and not in vscode, I'm not sure what could be up with that?

PR #1907, opened 2025-12-01
🩹 Fix sequence

SnaveSutit:
This change allows the `gapParser` to be used to show context errors, like "expected space", between arguments, while also allowing for optional arguments in a sequence with such a `gapParser`.

The current behavior is gap parser will run between every item in the sequence, whether or not those parsers actually moved the cursor or returned nodes. So multiple chained optional arguments are fiddley to implement.
```ts
function sequenceDemo(): core.Parser<core.AstNode> {
    // An InfallableParser that produces an error if it encounters a gap that isn't exaclty one space.
    const sep = core.map(mcf.sep, () => [])

    return core.setType(
        'sequence_demo',
        core.sequence([
            core.literal('sequence'),
            // gapParser is called
            core.optional(
                core.failOnEmpty(
                    core.resourceLocation({
                        category: 'tag/function',
                        usageType: 'reference',
                        allowTag: false,
                    }),
                ),
            ),
            // In the current implementation, the gapParser (sep) is called again between these
            // two parsers, whether or not the optional parser actually captured anything.
            // Because the gapParser shows an error if the gap isn't exactly one space,
            // it will always show an error when the optional parser returns `undefined`, as
            // the space has already been consumed by the previous gapParser call.
            core.literal('test'),
        ], sep),
    )
}
```
To work around this issue with the current parser, we need to put `sep` between every argument (tedious code duplication), and add an extra `core.sequence` call inside the optional parser in order to only check for that `sep` when it's actually needed creating extra tree depth that could be avoided. (adds more indentation levels. :nauseated_face:)
```ts
function sequenceDemo(): core.Parser<core.AstNode> {
    // An InfallableParser that produces an error if it encounters a gap that isn't exaclty one space.
    const sep = core.map(mcf.sep, () => undefined)

    return core.setType(
        'sequence_demo',
        core.sequence([
            core.literal('sequence'),
            sep,
            core.optional(
                core.sequence([
                    core.failOnEmpty(
                        core.resourceLocation({
                            category: 'tag/function',
                            usageType: 'reference',
                            allowTag: false,
                        }),
                    ),
                    sep,
                ]),
            ),
            core.literal('test'),
        ]),
    )
}
```

### DIFF
--- a/packages/core/src/parser/util.ts
+++ b/packages/core/src/parser/util.ts
@@ -104,25 +104,29 @@ export function sequence(
 			range: Range.create(src),
 		}
 
-		for (const [i, p] of parsers.entries()) {
+		for (const p of parsers) {
 			const parser = typeof p === 'function' ? p : p.get(ans)
 			if (parser === undefined) {
 				continue
 			}
 
-			if (i > 0 && parseGap) {
-				ans.children.push(...parseGap(src, ctx))
-			}
-
+			const start = src.cursor
 			const result = parser(src, ctx)
+
 			if (result === Failure) {
 				return Failure
-			} else if (result === undefined) {
-				continue
 			} else if (SequenceUtil.is(result)) {
 				ans.children.push(...result.children)
-			} else {
+			} else if (result !== undefined) {
 				ans.children.push(result)
+			}
+
+			if (
+				parseGap
+				// If the parser didn't move the cursor, and returns undefined, calling parseGap is redundant.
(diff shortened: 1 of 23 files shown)


MulverineX (changes requested):
Most of these updated snapshots are incorrect results

MulverineX:
afaik this PR has been superseded by more current work in Snave's full MCB fork branch

SnaveSutit:
Yep 👍

PR #1932, opened 2025-12-21
✨ Mcdoc bigint

NeunEinser:
Mcdoc long literals, and ranges on long types are now read as a bigint, retaining maximum precision.

A `toJSON` function is patched onto the BigInt class according to the best practices outlined here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#use_within_json and here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/BigInt_not_serializable#providing_a_custom_serialization_method

Per default, any bigint will be serialized as a raw JSON number. Consumers of the JSON need to read integers without precision loss in order to benefit from this. If this is not wanted and should e.g. be serialized as a JSON string instead, this should be done explicitly in this case.

The SymbolTable has been altered to be able to serialize and deserialize bigints on its data field consistently. For this purpose, during `unlink` all bigints in the data field will be replaced with an object of the following structure: `{"$$type": "bigint", "value": "123456789"}`.

When `link`ing an unlinked table (or a deserialized table), this replacement is undone again, retaining the original structure.

The SymbolTable serialization is important for our cache. If it stayed untouched, the precision would be lost by JS's JSON.parse implementation, resulting in no benefit after a cache write and read cycle.


packages/core/src/common/patches.ts
```diff
@@ -0,0 +1,5 @@
+export function applyPatches() {
+	;(<any> BigInt.prototype).toJSON = function() {
+		return (<any> JSON).rawJSON(this.toString())
+	}
```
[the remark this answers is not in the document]
NeunEinser: It is needed as a fallback for other sources.

In practice, for example mcdoc-cli would error without it, and imo it is good practice to just have a default fallback, rather than having to implement / respect it in all places.

That handling in snapshot-setup probably could just be removed entirely now.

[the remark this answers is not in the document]
SPGoding: Seems like we need to wait for this (or add it ourselves since it has the `Help Wanted` label?): https://redirect.github.com/microsoft/TypeScript/issues/61330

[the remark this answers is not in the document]
misode: Yeah, I am not a fan of prototype modifications like this

misode (changes requested):
I am not convinced that we need this. On the surface, using bigint for long ranges might seem cleaner and better, but the reality is that there is no clean way to implement this, while keeping mcdoc types serializable to JSON. I really dislike the special `$$type` JSON format and `rawJSON` patching.

I understand there is a benefit in using bigint for these, but I don't think that weighs up against all the disadvantages.

SPGoding:
> I really dislike the special `$$type` JSON format and `rawJSON` patching.

The `rawJSON` replacement of BigInts could technically be done in a `replacer` in `JSON.stringify()` so that no prototype patching is needed, and we could access the raw text through the [`reviver` argument of `JSON.parse()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#the_reviver_parameter) to restore the number literal in JSON back to BigInt. The challenge would be to know which number literals to convert to BitInts and which to keep as numbers in the reviver.

MulverineX:
> The challenge would be to know which number literals to convert to BitInts and which to keep as numbers in the reviver.

IMO we move all long literals to BigInt, and proceed with the `{"$$type": "bigint", "value": "123456789"}` serialization. Checking for bounds during the serialization step and the parse step is pretty annoying.

As for implementation path, I agree with SPG here, we should just use the built in `JSON` method arguments

NeunEinser:
> Why do we have the "lossless" replacer/reviver when we have these?

Have you read my comment above your review? If yes please explain what you do not understand about what I am saying there.

packages/mcdoc/src/type/index.ts
```diff
@@ -95,14 +102,17 @@ export namespace NumericRange {
...
 
-	export function intersect(a: NumericRange, b: NumericRange): NumericRange {
-		const min: number | undefined = a.min !== undefined && b.min !== undefined
-			? Math.max(a.min, b.min)
+	export function intersect<T extends (number | bigint) = number>(
+		a: NumericRange<T>,
+		b: NumericRange<T>,
+	): NumericRange<T> {
+		const min: T | undefined = a.min !== undefined && b.min !== undefined
+			? (a.min > b.min ? a.min : b.min)
```
SPGoding: (nit) can we add our own `min()` / `max()` functions somewhere (`core/src/common/util.ts` maybe?) with a `<T extends (number | bigint)>` generic?

packages/mcdoc/src/type/index.ts
```diff
@@ -95,14 +102,17 @@ export namespace NumericRange {
 			&& a.max === b.max
```
SPGoding: Is it intentional that a bigint range and a number range with equal numeric values are not considered equal?

packages/mcdoc/src/type/index.ts
```diff
@@ -490,7 +519,7 @@ export namespace McdocType {
 	}
 
 	export function toString(type: McdocType | undefined): string {
-		const rangeToString = (range: NumericRange | undefined): string => {
+		const rangeToString = (range: NumericRange<any> | undefined): string => {
```
SPGoding: What happens if we omit the `<any>` here so it uses the default `<number | bigint>`? And if that breaks something, could we use `unknown` instead of `any` here?

NeunEinser: I don't remember why I did this, but yes we can just omit it.

packages/mcdoc/src/type/index.ts
NeunEinser: This is now also technically wrong for the case in which `T` allows both `number | bigint`, for the case that either a.min === b.min or a.max === b.max, and b is exclusive while a is not.

packages/mcdoc/src/type/index.ts
NeunEinser: I think it makes sense to treat bigint and number the same here. For the current usages for sure, and I don't see a reason you'd ever need a strict equals here.

packages/core/test/common/json.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+	return JSON.stringify(obj, bigIntTestDescriptorReplacer)
+		.replace(/"(\d+n)"/g, '$1')
+		.replace(/"([a-zA-Z0-9_.-]+)":/g, '$1:')
+
+	function bigIntTestDescriptorReplacer(_key: string, value: any) {
+		return typeof value === 'bigint'
+			? `${value}n`
+			: value
+	}
+}
```
NeunEinser: I wonder if it might make sense to use this serialization in the `snapshot-setup` in `snapshot.setDefaultSnapshotSerializers` in general.

It will clearly show the type in a js object style literal (as you can see in the snapshot file of this test). I used it here since the snapshot would look very confusing when the types are not clear, but I don't think this would hurt for other snapshots in general either.

packages/core/test/common/json.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+		.replace(/"(\d+n)"/g, '$1')
+		// Remove quotes around keys if possible
+		.replace(/"([a-zA-Z0-9_.-]+)":/g, '$1:')
+
+	function bigIntTestDescriptorReplacer(_key: string, value: any) {
+		return typeof value === 'bigint'
+			? `${value}n`
+			: value
+	}
+}
```
NeunEinser, line 92: I wonder if it might make sense to use this serialization in the `snapshot-setup` in `snapshot.setDefaultSnapshotSerializers` in general.

It will clearly show the type in a js object style literal (as you can see in the snapshot file of this test). I used it here since the snapshot would look very confusing when the types are not clear, but I don't think this would hurt for other snapshots in general either.

SPGoding: You can try that. My main worry would be the two regexes matching something inside a string instead of just the outer key/values.

SPGoding (approved):
LGTM, thanks for this

packages/core/test/common/json.spec.ts
NeunEinser: What do you mean by outer key/values?

SPGoding: I was worrying what if hypothetically in one of our test case we have values shaped like `"(\d+n)"` or `"([a-zA-Z0-9_.-]+)":` inside a string, but now I realize that can't really happen as the quote would need to be escaped in those cases and the ~~regexes~~ ~~regices~~ regular expressions wouldn't match the backslash

MulverineX (approved):

PR #1933, opened 2025-12-21
✨ Strict enum

NeunEinser:
This enforces enums to only use the type they declare.

For backwards compatibility, explicit type suffixes are still possible, but they must match the enum's type.

Type suffixes are still optional, the only difference is that the correct type is now always enforced, even when the suffix is missing.

image
image
image
image
image


SPGoding:
The error messages seem a bit awkward to me (esp. the `Encountered unparseable content`) since we physically cannot parse those unexpected values. Do you think if it'd make sense to make this a binder-level error with a more appropriate error message instead?

SPGoding:
I agree the parser errors are not ideal but it is a challenging problem to solve. Even the TypeScript compiler freaks out if there's a missing ending parenthesis and gives rather useless errors.

image

I'm also not convinced having conditional syntax based on source text is a good idea. I'd like the syntax to stay static and corresponds to the spec, and have validation logic on a different layer (e.g. binder) instead.

MulverineX:
I think this should just be addressed with a binder-level error

SPGoding (approved):

MulverineX (approved):

Issue #1938, opened 2025-12-24
Alternative options for obtaining required meta-resources


SPGoding:
`@spyglassmc/java-edition` requires the following meta-resources to function (I use the word "meta-resources" to disambiguate it from "resources" from resource packs):

* `versions.json`: (shared by all Minecraft versions) the versions list from mcmeta. https://github.com/misode/mcmeta/blob/summary/versions/data.json
* `vanilla-datapack`: (unique for each Minecraft version) the `data` branch from mcmeta. https://github.com/misode/mcmeta/tree/data-json
* `vanilla-resourcepack`: (unique for each Minecraft version) the `assets-tiny` branch from mcmeta, unique for each Minecraft version. https://github.com/misode/mcmeta/tree/assets-tiny
* `vanilla-mcdoc`: (shared by all Minecraft versions) the vanilla-mcdoc source code. https://github.com/SpyglassMC/vanilla-mcdoc
* `mcmeta-summary`: (unique for each Minecraft version) includes `block_states.json`, `commands.json`, and `registries.json`. https://github.com/misode/mcmeta/tree/summary

Currently, all meta-resources are fetched from the `api.spyglassmc.com` API. Any missing meta-resource causes undefined behavior. This makes the extension unusable if `api.spyglassmc.com` is unreachable due to Internet restrictions (#1883), garbage server running garbage code (#1896), or user being offline.

Some alternative ways of obtaining such meta-resources will be explored.

### Alt 1: Bundled with package

The meta-resources can be bundled with the `@spyglassmc/java-edition` package in some fashion, details be damned, and used as a fallback.

* This is a very attractive option for the `versions.json` list and `vanilla-mcdoc`, as they (1) are shared by all Minecraft versions so they provide full backwards compatibility, and remain largely usable forwards even if slightly outdated; and (2) are relatively small in sizes so they won't increase the bundle size significantly (`versions.json` is `22.1kB` minimized and gzipped, `vanilla-mcdoc.tar.gz` is `124.1kB`; for reference the VS Code extension `vsix` package is `429.70kB`).
* For `mcmeta-summary`, they are still relatively small (`block_states.json` is `9.58kB`, `commands.json` is `7.46kB`, and `registries.json` is `116kB`, all minimized and gzipped), but are no longer version agnostic. Realistically we can bundle at most one for the latest release version and one for the latest snapshot version, and use some heuristics to pick the most applicable one for fallback.
[...]

PR #1940, opened 2025-12-24
🐛 Don't show “expected nothing” if root type cannot be resolved

NeunEinser:
Work towards #1938

> - We should have graceful fallback behaviors for any missing meta-resources.
>   - We would rather have no validations than false positives (full screens of nonsensical errors like "Expected nothing" is not acceptable)

With this change, in case the root type the mcdoc runtime checker is called with cannot be resolved, the `any` type will be used instead of empty union.

Usually in case something is not able to be resolved, an empty union is emitted, causing an “expected noting” error. In case this happens for the type runtime checker is called with, this will be replaced by the `any` type. Calling the runtime checker explicitly with an empty union will still produce “expected noting”, and in case a type that is referenced later cannot be resolved, “expected nothing” would still be emitted.

This disables all validations and completions for this case in a graceful way.

### DIFF
--- a/packages/mcdoc/src/runtime/checker/index.ts
+++ b/packages/mcdoc/src/runtime/checker/index.ts
@@ -206,6 +206,14 @@ export function typeDefinition<T>(
 		const validRootDefinitions = simplifiedRoot.kind === 'union'
 			? simplifiedRoot.members
 			: [simplifiedRoot]
+
+		if (
+			validRootDefinitions.length === 0
+			&& (typeDef.kind !== 'union' || typeDef.members.length > 0)
+		) {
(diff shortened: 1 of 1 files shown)


SPGoding (approved):
For reference the consensus is to only use `any` for the root type for now, and we can re-evaluate if it makes sense to do that for other types at a later point once we can ensure all missing reference errors in mcdoc files would be reported to the user so that those problems aren't completely hidden away with `any` types at validation time.

NeunEinser:
(The discussion SPG is referring to can be found here in Spyglass Discord for reference: https://discord.com/channels/666020457568403505/1498745476399366226)

PR #1942, opened 2025-12-26
♻️ Use JSON modules for locale JSON files

SPGoding:


packages/locales/package.json
```diff
@@ -7,21 +7,9 @@
...
-    "build": {
-      "command": "node ./build.js",
-      "files": [
-        "src/locales/*.json"
-      ],
-      "output": [
-        "lib/locales/*.js"
-      ]
-    }
-  },
```
SPGoding: A custom build step was used to copy JSON locale files as JavaScript files that can be imported. This is no longer needed with JSON modules.

packages/tsconfig-base.json
```diff
@@ -2,14 +2,14 @@
 	"compilerOptions": {
 		"allowSyntheticDefaultImports": true,
 		"esModuleInterop": true,
-		"module": "ES2022",
-		"moduleResolution": "Bundler",
+		"module": "node20",
+		"moduleResolution": "node16",
```
SPGoding: It's unclear to me why the `moduleResolution` was changed from `node16` to `Bundler` in https://github.com/SpyglassMC/Spyglass/pull/1121. The only difference between `node16` and `Bundler` is that `Bundler` allows omission of file extensions ([doc](https://www.typescriptlang.org/tsconfig/#moduleResolution)), which is incorrect for us because every package in our repo is first and foremost a package for running on Node.js, so they must conform to Node.js's module resolution requirements. The fact that some packages are later ran through a bundler (esbuild) for `playground` and `vscode-extension` is not relevant.

I've also updated `module` to `node20` following a similar reasoning that every package must run on Node.js. In practice it should behave the same as `ES20xx` as every single package has `"type": "module"` defined in their `package.json`.

PR #1953, opened 2026-01-21
✨ Support new SNBT Unicode escapes from 25w09a

### LINKED ISSUE #1771: Support 25w09a SNBT syntax
https://www.minecraft.net/en-us/article/minecraft-snapshot-25w09a

* Allow `.1` and `1.` numbers
* Allow float numbers with `e` notation
* Allow hex and binary numbers starting with `0x` and `0b`
* Enforce that numbers can't start with `0`
* Allow numbers to have `_` separators
* Allow `s` and `u` (signed and unsigned) suffixes
* String escape sequences
* Disallow unquoted strings starting with `[0-9.+-]`
* Allow trailing commas in collections and compounds

Calverin:
Adds support in SNBT Strings for:
- 2 digit Unicode sequences - `\x0A`
- 8 digit Unicode sequences - `\U0001F525`
- Named Unicode characters - `\N{Fire}` (validated against [this list](https://unicode.org/Public/UNIDATA/UnicodeData.txt) with control characters using their secondary name and any character with parentheses in the name being omitted)

Checks a box in #1771


MulverineX (approved):

misode (changes requested):

SPGoding (commented):
Can we add some test cases for these new escape sequences? A few cases off the top of my mind:
1. Using them when `extendedUnicode` is disabled
2. Using them when `extendedUnicode` is enabled
3. Testing various syntax errors you can have with them

And would it be possible to commit the script you used to generate the lookup table as well so we can re-run it in the future? I would prefer it if it's put under somewhere like `packages/core/scripts` to keep it separate from the production source code but if the script is in TypeScript you might need to update `packages/core/tsconfig.json` to add a project reference to it along with `src` and `test` and create a new `packages/core/scripts/tsconfig.json` (just putting `{ "extends": "../../tsconfig-type-strip" }` as its content should ba fine).

packages/core/src/parser/string.ts
```diff
@@ -52,7 +59,52 @@ export function string(options: StringOptions): InfallibleParser<StringNode> {
...
+							)
+							ans.valueMap.push({
+								inner: Range.create(ans.value.length, ans.value.length + 1),
+								outer: Range.create(cStart, src),
+							})
+							ans.value += c2
+							cStart = src.cursor
+							continue
+						}
+						const name = src.peekUntil('}')
```
misode: Minecraft actually allows whitespace between the curly braces and the name. So you might need to trim the name or add some `src.skipSpace` calls.

misode (changes requested):
Given some of the discussion on discord, are we now sure that this JSON file 100% matches what the vanilla game accepts?

NeunEinser:
> Given some of the discussion on discord, are we now sure that this JSON file 100% matches what the vanilla game accepts?

I think we should do some code analysis (maybe of the JDK?) to see which list it uses, and how it uses that list. Some dependencies have been found.

Given the complexity these names cause, imo it would also be acceptable for this PR to either not handle them, or allow anything in there without giving validation or completion.

It would be very nice to get this merged in general.

packages/core/src/parser/string.ts
NeunEinser: imo trimming is probably better -- you can't just skip any space as the names can also contain spaces. Dealing with this using skipSpace sounds very complicated

packages/core/src/parser/string.ts
```diff
@@ -52,7 +59,52 @@ export function string(options: StringOptions): InfallibleParser<StringNode> {
...
+						} else if (
+							/^[-a-zA-Z0-9 ]+$/.test(name) && UnicodeLookupTable.has(name.toLowerCase())
+						) {
+							src.skip(name.length + 1)
+							ans.valueMap.push({
+								inner: Range.create(ans.value.length, ans.value.length + 1),
+								outer: Range.create(cStart, src),
+							})
+							ans.value += String.fromCodePoint(UnicodeLookupTable.get(name.toLowerCase())!)
+						} else {
```
NeunEinser, line 153: This is not handling all cases 100% correctly.

- Check if the character matches the pattern `/^([a-z0-9 -]+) ([a-f1-9][a-f0-9]+)$/i`. Interpret the first matching group as the code block and the second matching group as the code block
  - Check that the given codepoint does not map to a character string code (e.g. use an inverse map of the UnicodeLookupTable for this)
  - Check that the code block is a valid code block according to [Blocks.txt](https://unicode.org/Public/UNIDATA/Blocks.txt)
  - Check that the codepoint lies between a `<..., First>` and `<..., Last>` according to [UnicodeData.txt](https://unicode.org/Public/UNIDATA/UnicodeData.txt) OR that this is a character explicitly listed in this file, and the name is either empty; or the name is `<control>` and the secondary name is also empty. You currently do not include these empty names in your json file (could be mapped to `null`, I suppose). Note that the `<..., First>` and `<..., Last>`  are also not present in the JSON file you provide.

packages/core/src/parser/unicode-lookup-table.json
NeunEinser, line 1: There are four special cases:

- `0x0007` is `BEL` (`BELL` is still valid but maps to `1F514`; your json lookup does handle map the `BELL` character to the correct codepoint already, though it is listed earlier than normal due to this anomoly)
- `0x0080` is `PADDING CHARACTER` (this would be unnamed otherwise)
- `0x0081` is `HIGH OCTET PRESET`
- `0x0099` is `SINGLE GRAPHIC CHARACTER INTRODUCER`

You could either hardcode these in code, but also just modifying this json file would be fine.

NeunEinser (changes requested):
Credit to @Papierkorb2292 for the research.

It would also be fine to only validate that the text within the braces matches `/^[-a-zA-Z0-9 ]+$` and not validate / complete the names themselves.

For future work I also see:
- Make sure the correct UnicodeData.txt for the version of Minecraft is used
  - We probably would want to bundle a lookup-table with versioning data instead (so the string name can map to a list of objects in a format like this: `[{ "since": "69.0", "until": "69.2", "codepoint": 91},{"since":"69.2","codepoint":420}]`
- Add some build script (maybe sth like `npm run update-unicode-data`) to automatically update this file

Though those two points imo do not need to be addressed in this PR.

MulverineX:
I'm taking up the mantle on this PR, after getting it done I'm gonna work on all of the rest of https://github.com/SpyglassMC/Spyglass/issues/1771

MulverineX:
That commit is only Part 1, I've still got more work to do

MulverineX:
image

image

Papierkorb2292:
I think the special cases like `BEL` still aren't accepted, is the `JdkNameOverrides` table used? Also some of the names in that table should already be covered by UnicodeData.txt like `null` or `backspace`. Additionally `tab`, `newline`, and `carriage return` are not actually accepted by Java's `codePointOf` so I don't think they should be there either.

MulverineX:
Ah yep, a handful of those were slop additions, and `JdkNameOverrides` didn't actually get wired in, tested in-game and in-editor, should be good now. Sorry about that, thanks Papier

MulverineX:
[Screencast_20260809_143933.webm](https://github.com/user-attachments/assets/f69e47c7-180c-4970-977a-f490013e6389)

.github/workflows/release-legacy-v3.yml
```diff
@@ -0,0 +1,31 @@
+name: Release legacy-v3
```
misode: I don't think you merged correctly, because this workflow was removed on `main` and should definitely not be re-added.

packages/web-api-server/src/unicode.ts
```diff
@@ -0,0 +1,387 @@
...
+ * Conditional fetch using `Last-Modified`. Writes the response body to
+ * `dest` and stores the upstream `Last-Modified` timestamp in
+ * `${dest}.lastmod` for subsequent runs. Returns the (possibly freshly
+ * fetched) file contents.
+ *
+ * We deliberately use `Last-Modified` rather than ETag because
+ * `unicode.org`'s CDN does not honour weak ETags consistently across
+ * requests (gzip representation differences cause mismatches).
+ */
+async function fetchAndCache(
```
misode: Is there a reason you're not using the `fetchWithCache` core function?

.eslintrc.cjs
```diff
@@ -0,0 +1,99 @@
+module.exports = {
```
misode: This PR should not be adding an eslint config.

scripts/set_tsconfig_references.ts
```diff
@@ -0,0 +1,47 @@
+import * as fs from 'fs'
```
misode: This file was removed on main, why is it being re-added here?

scripts/refresh_unicode_data.ts
```diff
@@ -0,0 +1,315 @@
...
+	blocks: { [name: string]: [number, number] }
+}
+
+interface ParsedEntry {
+	codepoint: number
+	primary: string
+	secondary: string
+}
+
+function parseUnicodeDataEntries(text: string): ParsedEntry[] {
```
misode, line 39: A large part of the code in this file is duplicated also in `web-api-server/src/unicode.ts`.

packages/core/src/tsconfig.json
```diff
@@ -3,9 +3,13 @@
...
+		"**/*.ts",
+		"dependency/unicode-lookup-table.json"
+	],
 	"references": [
 		{
 			"path": "../../locales/src"
 		}
 	]
-}
+}
```
misode: Removes the final newline

packages/core/src/service/Context.ts
```diff
@@ -42,6 +42,7 @@ export interface ParserContext extends ContextBase {
 	config: Config
 	doc: TextDocument
 	err: ErrorReporter
+	symbols: SymbolUtil
```
misode: Parsers should not have access to the symbol table. I'm pretty sure this was a strict requirement, but SPGoding might be able to explain why.

misode (changes requested):

packages/core/src/service/Context.ts
SPGoding: Idea was parsing for different files could be potentially parallelized. Adding access to the global symbol table defeats that.

MulverineX: Ah okay yeah that's totally fair. Alright I'll refactor stuff into the checker step

MulverineX:
Edit: Outdated

Papierkorb2292:
I have no idea why, but for some reason the [attached file](https://github.com/user-attachments/files/31112811/escapeCrash.txt) crashes the language server for me with error message

```
/home/papierkorb2292/Documents/Spyglass/packages/core/src/checker/string.ts:340
	child.resolved = String.fromCodePoint(codepoint)
	                        ^

RangeError: Invalid code point NaN
    at String.fromCodePoint (<anonymous>)
    at finalizeEscape (/home/papierkorb2292/Documents/Spyglass/packages/core/src/checker/string.ts:340:26)
    at visit (/home/papierkorb2292/Documents/Spyglass/packages/core/src/checker/string.ts:379:6)
    at string (/home/papierkorb2292/Documents/Spyglass/packages/core/src/checker/string.ts:390:2)
    at <anonymous> (/home/papierkorb2292/Documents/Spyglass/packages/core/src/processor/checker/builtin.ts:91:19)
    at traversePreOrderImpl (/home/papierkorb2292/Documents/Spyglass/packages/core/src/processor/util.ts:35:3)
    at traversePreOrder (/home/papierkorb2292/Documents/Spyglass/packages/core/src/processor/util.ts:24:2)
    at Object.fallbackSync2 (/home/papierkorb2292/Documents/Spyglass/packages/core/src/processor/checker/builtin.ts:85:2)
    at <anonymous> (/home/papierkorb2292/Documents/Spyglass/packages/nbt/src/checker/index.ts:91:16)
    at Object.typed2 (/home/papierkorb2292/Documents/Spyglass/packages/nbt/src/checker/index.ts:23:32)
```
It's very weird, it only seems to happen if there's a multiline thing before the command. So if I remove the  `#\` at the start of the file it doesn't crash anymore. Same thing happens with multiline commands or macros.

MulverineX:
I'll try to figure out why it has issues when multiline is involved

MulverineX:
Resolved, thanks

packages/core/src/checker/string.ts
```diff
@@ -0,0 +1,385 @@
...
+	child: UnicodeEscapeNode,
+	ctx: CheckerContext,
+	node: StringBaseNode,
+): void {
+	const raw = child.raw
+	const escapeRange = child.range
+	// Reject extended escapes (\u/\U/\x/\N{...}) when the loaded/resolved
+	// game version predates 1.20.5. The parser accepted them syntactically;
+	// the checker is responsible for the version-gated diagnostic. Source of
+	// truth: `project.ctx.loadedVersion` (set by `je.initialize`), with
```
SPGoding: `project.ctx.loadedVersion` is a java-edition layer concept. Does the version check belong in core?

imo the version check should happen in java-edition. Checker logic can stay in core and gated by a flag, or be moved to java-edition as well.

packages/java-edition/src/dependency/unicode.ts
```diff
@@ -0,0 +1,282 @@
...
+					data: { data: data.ranges },
+					usage: { type: 'definition' },
+				})
+			symbols.query(BlocksUri, UnicodeBulkCategory, BulkBlocks)
+				.enter({
+					data: { data: data.blocks },
+					usage: { type: 'definition' },
+				})
+		})
+	}
```
SPGoding, line 224: How long does this registrar take to run? And how much space does all the data take up in the symbol table?

packages/core/src/processor/completer/builtin.ts
```diff
@@ -260,11 +260,23 @@ export const resourceLocation: Completer<ResourceLocationNode> = (node, ctx) =>
...
+		if (child) {
+			return dispatch(child, ctx).map(item => ({
+				...item,
+				filterText: escapeString(item.filterText ?? item.label, node.quote),
+				insertText: escapeString(item.insertText ?? item.label, node.quote),
+			}))
+		}
+	}
+
+	const unicodeNameItem = tryGetUnicodeNameCompletion(node, ctx)
```
SPGoding: Should unicode name completion be merged with value node completion instead? Right now I understood it as we don't get unicode name completions if the string has a value parser

packages/core/src/parser/string.ts
```diff
@@ -17,6 +40,7 @@ export function string(options: StringOptions): InfallibleParser<StringNode> {
 			options,
 			value: '',
 			valueMap: [],
+			children: [],
```
SPGoding: Can we only create this when we need to

package.json
```diff
@@ -22,6 +22,8 @@
 		"docs:start": "cd docs && (bundle exec jekyll serve --incremental || true) && cd ..",
 		"fmt": "dprint fmt",
 		"fmt:check": "dprint check",
+		"refresh-unicode-data": "node scripts/refresh_unicode_data.ts",
+		"refresh-unicode-data:force": "node scripts/refresh_unicode_data.ts --force",
```
SPGoding: (nit) can we sort the scripts alphabetically

packages/core/src/checker/string.ts
MulverineX: image

We're already doing it, figured it was fine, do you need me to address this before we get the PR merged?

packages/core/src/parser/string.ts
MulverineX: I thought the standard in our LS was to keep an empty array rather than omit a key, but yeah I can do that

packages/core/src/processor/completer/builtin.ts
MulverineX: Can you give an example command? I'm not sure what you mean here

packages/java-edition/src/dependency/unicode.ts
MulverineX: Speed:

Added logging for all registrars

image

If you want I can make `unicode-lookup-table.json` exactly the same as what's needed for the symbol table, that might increase performance a little?

MulverineX: Size: `7.46KiB` (gzipped)

Without unicode registrar: `1.3 MiB (1,364,026)`

With unicode registrar: `1.3 MiB (1,371,485)`

(checked using `/var/home/mulverine/.cache/spyglassmc-nodejs/symbols/_.json.gz`)

packages/core/src/checker/string.ts
SPGoding: Well Misode did also leave a `TODO` saying not to do that :)
I think the problem is also worse in `core` than in `nbt` as well so I'd like to see it addressed.

packages/java-edition/src/dependency/unicode.ts
SPGoding: Thanks for checking! This seems fine

packages/core/src/processor/completer/builtin.ts
SPGoding: Basically when a string has a value parser (e.g. resource locations in this case),

```mcfunction
summon minecart ~ ~ ~ {DisplayState: {Name: ""}}
```

image

Unicode names are no longer suggested:

image

MulverineX: image

image

image

Yeah I guess this is kinda jank, auto complete of resource locations *does* work though

MulverineX: Eh this'd be really hard to fix unless I'm missing something, do we need to do this for the PR or can it just be added to the backlog?

(Also I'm not really sure what your desired fix is here)

SPGoding: Does something like this not work https://github.com/SpyglassMC/Spyglass/commit/b5a5f1ea8cb21a2531ae8c1661c271582383bef4

If it's too much work then I guess this is fine

SPGoding (approved):

packages/core/src/processor/completer/builtin.ts
MulverineX: Oh that totally works, just tested it on top of this PR, thanks, applying that commit

packages/core/src/processor/checker/builtin.ts
```diff
@@ -121,6 +121,6 @@ export const symbol: Checker<SymbolBaseNode> = (_node, _ctx) => {
 }
 
 export function registerCheckers(meta: MetaRegistry) {
-	meta.registerChecker<ResourceLocationNode>('resource_location', resourceLocation)
-	meta.registerChecker<SymbolNode>('symbol', symbol)
+	// meta.registerChecker<ResourceLocationNode>('resource_location', resourceLocation)
+	// meta.registerChecker<SymbolNode>('symbol', symbol)
```
NeunEinser, line 125: Is this intentional?

packages/core/src/processor/colorizer/builtin.ts
```diff
@@ -78,20 +79,59 @@ export const resourceLocation: Colorizer<ResourceLocationBaseNode> = (node, _ctx
...
+			ColorToken.create(
+				Range.create(range.start + 3, range.end - 1),
+				'resourceLocation',
+			),
+		)
+		tokens.push(
+			ColorToken.create(Range.create(range.end - 1, range.end), 'escape'),
+		)
+	} else {
+		// `\xHH` / `\uHHHH` / `\UHHHHHHHH`: hex digits
```
NeunEinser, line 127: Uhhhhh 🤣

packages/java-edition/src/checker/index.ts
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Appends `extra` to the checker already registered for `type`, instead of
+ * replacing it. `MetaRegistry.registerChecker` is a plain map write, so a
+ * naive registration here would silently drop the checker that another
+ * package registered for the same node type.
+ *
+ * The result stays synchronous when the existing checker is synchronous:
+ * `checker.fallbackSync` warns about checkers that return a promise.
+ */
```
NeunEinser, line 33: Maybe sth like this could also work for the NBT stuff on the other PR, rather than adding that field to the core MetaRegistry

packages/json/src/parser/string.ts
```diff
@@ -2,7 +2,10 @@ import * as core from '@spyglassmc/core'
 import type { JsonStringNode } from '../node/index.js'
 
 export const JsonStringOptions: core.StringOptions = {
-	escapable: { characters: ['b', 'f', 'n', 'r', 't'], unicode: true },
+	// Always accept the full escape syntax; the edition package's string
+	// checker reports version-gated errors when the loaded game version
+	// predates the support cutoff.
+	escapable: { characters: ['b', 'f', 'n', 'r', 't'], unicode: true, extendedUnicode: true },
```
NeunEinser: Isn't this wrong? JSON escapes don't support all of this, or?

packages/java-edition/src/checker/string.ts
```diff
@@ -0,0 +1,370 @@
...
+			const children = n.children ?? []
+			const isStringNode = (n as StringBaseNode).options?.escapable !== undefined
+			if (isStringNode) {
+				for (const child of children) {
+					if (child.type === 'unicode_escape') {
+						finalizeEscape(
+							child as UnicodeEscapeNode,
+							ctx,
+							n as StringBaseNode,
+							supportsExtendedEscapes,
```
NeunEinser, line 336: Don't you need to consider the node's options.escapable here, to know what the node itself allows? And if I am just misunderstanding how you are handling these sequences not being allowed in other conexts like JSON, why does StringOptions need this property at all?

packages/java-edition/src/checker/string.ts
```diff
@@ -0,0 +1,370 @@
...
+	isUnicodeNamesByCodepointMap,
+	isUnicodeRangeMap,
+	ReleaseVersion,
+	toTitleCase,
+	UnicodeBulkCategory,
+	UnicodeDataUri,
+	UnicodeNameCategory,
+} from '../dependency/index.js'
+
+const NamedEscapeWithHexPattern = /^\s*([a-z0-9-]+(?: [a-z0-9-]+)*)\s+([a-f0-9]+)\s*$/i
```
NeunEinser: This regex allows for e.g. `\N{Hangul Syllables     AC00` with multiple spaces before the hex code.

This is not accepted by the game though:
Image

(Note that the command would actually work in chat, because chat converts multiple spaces into a single space)

packages/java-edition/src/checker/string.ts
```diff
@@ -0,0 +1,370 @@
...
+): number | undefined {
+	const query = ctx.symbols.query(
+		UnicodeDataUri,
+		UnicodeNameCategory,
+		`${toTitleCase(name)} `,
+	).symbol ?? ctx.symbols.query(
+		UnicodeDataUri,
+		UnicodeNameCategory,
+		toTitleCase(name),
+	).symbol
```
NeunEinser: When can it happen that the category name registered into the symbol table has a trailing space? Can we prevent that?

packages/java-edition/src/checker/string.ts
```diff
@@ -0,0 +1,370 @@
...
+					'parser.string.hex-expected',
+					rangeProbe[1]!,
+					start.toString(16).toUpperCase(),
+					end.toString(16).toUpperCase(),
+				),
+				innerRange,
+			)
+		}
+	}
+	const trailingMatch = TrailingNamePattern.exec(escape)
```
NeunEinser: Can you explain in words what trailing name pattern is, and give an example for it?

I also didn't see a test that references this pattern.

packages/java-edition/src/checker/string.ts
NeunEinser: When can it happen that the category name registered into the symbol table has a trailing space? Can we prevent that?

packages/core/src/processor/checker/builtin.ts
MulverineX: Yes, said checkers are no-ops and were no-ops before my PR so they don't need to get registered

packages/java-edition/src/checker/string.ts
MulverineX: This is intentional, I added the trailing space so accepting the completion for something like `Hangul Syllables ` drops the user directly into the hex codepoint slot. AFAICT there wasn't any other easy way to accomplish this.

packages/java-edition/src/checker/string.ts
MulverineX: "why does JSON not have these sequences": the parser rejects them (now). A JSON string AST will never carry one, so no further escape-mode check is needed in this checker

As for the rest, there should've been an assertion, whoops.

MulverineX:
Waiting on highlighting opinion from SPG and Papier in the discord

packages/java-edition/src/checker/nbtSyntax.ts
```diff
@@ -0,0 +1,229 @@
...
+	meta.registerChecker<NbtDoubleNode>('nbt:double', checkDouble)
+	meta.registerChecker<NbtStringNode>('nbt:string', checkString)
+	meta.registerChecker<NbtCompoundNode>('nbt:compound', checkCompound)
+	meta.registerChecker<NbtListNode>('nbt:list', checkList)
+	meta.registerChecker<NbtByteArrayNode>('nbt:byte_array', checkByteArray)
+	meta.registerChecker<NbtIntArrayNode>('nbt:int_array', checkIntArray)
+	meta.registerChecker<NbtLongArrayNode>('nbt:long_array', checkLongArray)
+
+	meta.registerChecker<core.StringNode>('string', unicodeEscapes)
+	meta.registerChecker<core.StringBaseNode>('json:string', unicodeEscapes)
```
NeunEinser: Should these two really be in a file called `nbtSyntax`?

packages/java-edition/src/checker/string.ts
```diff
@@ -154,99 +127,106 @@ function resolveNamedEscape(
...
+	const codepointAsFull = lookupName(inner, ctx)
+	if (codepointAsFull !== undefined) {
+		if (!isInDeclaredBlock(codepointAsFull, ctx)) {
+			ctx.err.report(
+				localize('parser.string.illegal-unicode-escape-name'),
+				escapeRange,
+			)
+			return undefined
+		}
+		if (!isValidUnicodeCodepoint(codepointAsFull, ctx)) {
```
NeunEinser: Do we need this check here? If we can already resolve the character name, could there ever be a case where a named unicode character we can map to a character code is not valid somehow?

packages/java-edition/src/checker/string.ts
```diff
@@ -154,99 +127,106 @@ function resolveNamedEscape(
...
 	}
-	if (!isInDeclaredBlock(codepoint, ctx)) {
-		ctx.err.report(
-			localize('parser.string.illegal-unicode-escape-name'),
-			escapeRange,
-		)
-		return undefined
+	const codepointAsFull = lookupName(inner, ctx)
+	if (codepointAsFull !== undefined) {
+		if (!isInDeclaredBlock(codepointAsFull, ctx)) {
```
NeunEinser: This is check already exists as an inlined version in the `isValidUnicodeCodepoint`, so this should definitely not be necessary.

packages/java-edition/src/checker/nbt/string.ts
```diff
@@ -154,99 +127,106 @@ function resolveNamedEscape(
...
 	}
-	if (!isValidUnicodeCodepoint(codepoint, ctx)) {
+
+	// 3. Whole-string name didn't resolve — try splitting: last token is the
+	// tail, everything before is the name. Handles `\N{Hangul Syllables D800}`
+	// (valid range + hex), `\N{Hangul Syllables FFFFF}` (range + out-of-range
+	// hex), and `\N{Hangul Syllables garbage}` (range + non-hex tail).
+	const words = inner.split(/\s+/)
+	if (words.length < 2) {
+		return reportIllegalName(ctx, escapeRange)
```
NeunEinser, line 145: (opt) It could be nice to maybe check first if this is a valid block name, and if so report sth like “Expected a character hex code in the range of xxxx - xxxx” only at the end of the name.

packages/java-edition/src/checker/string.ts
NeunEinser: I don't see any usage read of `extendedUnicode` of the StringOptions. So if the json parser already handles it somewhere, is this field needed at all?

packages/java-edition/src/checker/nbtSyntax.ts
MulverineX: Going with `nbt.ts` instead and I updated the `index.ts`:
```ts
import type * as core from '@spyglassmc/core'
import { registerNBT } from './nbt.js'
import { unicodeEscapes } from './string.js'

export function register(meta: core.MetaRegistry) {
	registerNBT(meta)

	meta.registerChecker<core.StringNode>('string', unicodeEscapes)
	meta.registerChecker<core.StringBaseNode>('json:string', unicodeEscapes)
}

export * from './nbt.js'
export * from './string.js'
```

packages/java-edition/src/checker/string.ts
MulverineX: Dropping this code, it was leftover from when the unicode mappings were loaded dynamically

packages/java-edition/src/checker/string.ts
MulverineX: Dropping this code, it was leftover from when the unicode mappings were loaded dynamically

packages/java-edition/src/checker/nbt/string.ts
MulverineX: Already an error that I just tested in-editor is being reported correctly, I'll try to make the code easier to read:
```
Hex codepoint expected after `Hangul Syllables`. Range: `0xAC00`-`0xD7A3`
```

packages/java-edition/src/checker/nbtSyntax.ts
misode: Are these complex escape sequences allowed in every `string` and `json:string` AST node? That seems like a lot.

packages/java-edition/src/checker/string.ts
MulverineX: Nope, in previous refactors it was made unnecessary, my bad for not cleaning it up, thanks!

NeunEinser (approved):

packages/java-edition/src/checker/nbtSyntax.ts
MulverineX: Ended up removing those checkers, they aren't needed, tested in-editor

Issue #1812, opened 2025-05-14
Firework explosion colors show an error for an int array, when both int lists and int arrays should be allowed


SomeYellowGuy:
The commit referencing #1783 only allows a list of `int`s, but should allow both an `int` array and list. I noticed this when I tried to use an MCStacker command. The issue specified that both types should be allowed, but the commit limited the allowed types to one.

Just to be sure, I checked the code, and both types are mentioned in the `xmap` function of the `Codec` and `PacketCodec` of firework colors in the `FireworkExplosionComponent` class, at least in 1.21.5.

![Image](https://github.com/user-attachments/assets/035c02eb-8850-4f35-9072-8ba1305c34fc)
![Image](https://github.com/user-attachments/assets/e538d1f8-6b00-4680-9b55-19e65a9350de)

NeunEinser:
This is not a concern of vanilla-mcdoc. The "native" type has been changed to a list of ints in 1.21.5, and this is what mcdoc is documenting.

The issue of not being lenient with types were Minecraft allows it is #1471

Issue #1500, opened 2024-07-09
"Expected a list" error at list


Bill2439:
Throwing "Expected a list" error on this: [I;0,0,0], which I am using to modify the Motion value of entities, works fine in game, throws an error in the editor.

Nico314159:
You're supposed to remove the `I;` part.

MulverineX:
Duplicate of #1471

Issue #1471, opened 2024-07-03
issues with using a number in boolean spaces for mcfunction files


LeonardsWorld:
this command works fine in game, but has an issue in VSC.
`/give @s minecraft:nether_star[minecraft:enchantment_glint_override=0]`
"enchantment_glint_override" allows both boolean values and simple numbers, so the correction saying that it must be boolean is false.

tryashtar:
For some more info, it appears to be SNBT, so `true` is just `1b`, and even `5e99f` can coerce its way to a bool here.

misode:
We're deciding to be strict here, just like other SNBT places in the extension. The components wants a boolean so there is no reason for the extension to allow integers, even if the game might implicitly convert.

SPGoding:
> The components wants a boolean so there is no reason for the extension to allow integers, even if the game might implicitly convert.

That will be a regression from v3. v3 had a lean SNBT checking mode that allows integers be used in place of a more correct type if `isPredicate` mode is not active.

misode:
Reopening. We'll need to change `isEquivalent` in the NBT package

NeunEinser:
I think it would be good to disallow explicitly declared integer types from casting to a different type though (e.g. 1b should warn when assigning to int, true should warn when used in a context where a regukar byte value is expected, etc)

There isn't currently a way for the isEquivalent checker to overwrite the error level with a warning though.

Nico314159:
I feel like that rule should be configurable. What I liked about the v3 linter was you could turn on/off and change the level of practically every rule.

SPGoding:
> I feel like that rule should be configurable.

The linter config for NBT type checking already exists: https://github.com/SpyglassMC/Spyglass/blob/e543fe70c2d9c743567a6d654d12527e2e9e2405/packages/core/src/service/Config.ts#L228

I don't know how much of the linter config is implemented though.

Nico314159:
> > I feel like that rule should be configurable.
> 
> 
> 
> The linter config for NBT type checking already exists: https://github.com/SpyglassMC/Spyglass/blob/e543fe70c2d9c743567a6d654d12527e2e9e2405/packages/core/src/service/Config.ts#L228
> 
> 
> 
> I don't know how much of the linter config is implemented though.

Specifically "boolean vs int strictness" should be its own subrule separate from the master toggle for type checking.

PR #1997, opened 2026-04-04
🐛 Replace isEquivalent with tryConvertTo

### LINKED ISSUE #2022: resource pack: axis rotation gives false error
A rotation in a resource pack, found in
- `blockstates/<name>.json` > `variants.<variant>.x` (and `.y`, `.z`) and
- `models/<path>/<name>.json` > `elements[].faces.<face>.rotation`,

currently gives an error saying the (valid) rotation is invalid.

for example (`blockstate`):
```json
{
	"variants": {
		"axis=x": {
			"model": "minecraft:block/acacia_log",
			"x": 90,
			"y": 90
		},
		"axis=y": {
			"model": "minecraft:block/acacia_log"
		},
		"axis=z": {
			"model": "minecraft:block/acacia_log",
			"x": 90
		}
	}
}
```
Image

https://github.com/SpyglassMC/vanilla-mcdoc/blob/90acdfb868ed95af297bc2a0b1c7054fc69bfb94/java/assets/block_state_definition.mcdoc#L29

NeunEinser:
This allows to convert a runtime type to a requested definition type.

Fixes #1991
Fixes #2022

Also modified the /nbt package to allow conversion between list types (when we are in non-canonical cases), working towards #1471 (effectively fixing the #1500 / #1812 / #1995 part of it)

### DIFF
--- a/packages/json/src/checker/index.ts
+++ b/packages/json/src/checker/index.ts
@@ -24,25 +24,36 @@ export function index(
 			[{ originalNode: node, inferredType: inferType(node) }],
 			type,
 			mcdoc.runtime.checker.McdocCheckerContext.create(ctx, {
-				isEquivalent: (inferred, def) => {
-					switch (inferred.kind) {
-						case 'list':
-							return ['list', 'byte_array', 'int_array', 'long_array', 'tuple'].includes(
-								def.kind,
+				tryConvertTo: (node, target) => {
+					switch (node.type) {
+						case 'json:array':
+							switch (target.kind) {
+								case 'list':
+									return { kind: 'list', item: { kind: 'any' } }
+								case 'byte_array':
+								case 'int_array':
+								case 'long_array':
+									return { kind: target.kind }
+								case 'tuple':
+									return { kind: 'tuple', items: [] }
+							}
+							break
+						case 'json:number':
+							const literalValue = mcdoc.LiteralNumericValue.makeIfValid(
+								target.kind,
+								node.value.value,
+								true,
+								true,
 							)
-						case 'struct':
-							return def.kind === 'struct'
-						case 'byte':
-						case 'short':
-						case 'int':
-						case 'long':
-							return ['byte', 'short', 'int', 'long', 'float', 'double'].includes(def.kind)
-						case 'float':
-						case 'double':
-							return ['float', 'double'].includes(def.kind)
-						default:
-							return false
+							if (literalValue !== undefined) {
+								return { kind: 'literal', value: literalValue }
+							}
 					}
+					const inferred = inferType(node)
+					if (inferred.kind === target.kind) {
+						return inferred
+					}
+					return undefined
 				},
 				getChildren: node => {
 					if (node.type === 'json:array') {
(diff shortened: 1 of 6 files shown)


packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -580,21 +592,33 @@ function checkShallowly<T>(
...
 						) {
 							foundMatch = true
 							otherKvpMatches.push(kvp.index)
 						}
 					}
 					for (const kvp of literalKvps.entries()) {
+						const literalType = {
+							kind: 'literal',
+							value: { kind: 'string', value: kvp[0] },
+						} satisfies LiteralType
```
misode: I think it's better to do `const literalType: LiteralType` here instead of using `satisfies`, so that the type of `literalType` is actually `LiteralType`. `satisfies` is only where you don't want to change the actual type of the result, but only want to check it.

packages/mcdoc/src/type/index.ts
```diff
@@ -2,6 +2,7 @@ import type { FullResourceLocation } from '@spyglassmc/core'
 import { Arrayable, Dev, max, min, numericEquals } from '@spyglassmc/core'
 import type { EnumKind } from '../node/index.js'
 import { getRangeDelimiter, RangeKind } from '../node/index.js'
+import { numericType } from '../parser/index.js'
```
misode: Unused import

packages/json/src/checker/index.ts
```diff
@@ -24,25 +24,36 @@ export function index(
...
-						default:
-							return false
+							if (literalValue !== undefined) {
+								return { kind: 'literal', value: literalValue }
+							}
+					}
+					const inferred = inferType(node)
+					if (inferred.kind === target.kind) {
+						return inferred
 					}
```
misode: Is this necessary in both the json and nbt packages? I would've expected that if the inferred kind is the same as the target kind then the runtime wouldn't need to try to call `tryConvertTo`.

packages/nbt/src/checker/index.ts
```diff
@@ -1,13 +1,21 @@
...
 	NbtCompoundNode,
+	NbtIntegerAlikeNode,
+	NbtListNode,
+	NbtLongNode,
 	NbtNode,
+	NbtNumberNode,
 	NbtPathFilterNode,
 	NbtPathIndexNode,
 	NbtPathKeyNode,
+	NbtPrimitiveArrayNode,
```
misode: Some unused imports in this file

misode (commented):
I'm going to be honest, it's been a while since I've really understood the runtime code. But the changes here look sensible and I can confirm it fixes the linked issue. Some nits/questions in review comments.

packages/json/src/checker/index.ts
NeunEinser: I struggle to remember the reason I did this, I don't think it should be needed? I think it might have been a safety thing, but calling `inferType` just for that seems strange.

misode (approved):

PR #2015, opened 2026-05-10
⬆️ TypeScript 6 7

SPGoding:
Still using tsc@6 for now since tsgo has different project reference loading order: https://github.com/microsoft/typescript-go/issues/3795

### DIFF
--- a/packages/tsconfig-base.json
+++ b/packages/tsconfig-base.json
@@ -4,12 +4,11 @@
 		"esModuleInterop": true,
 		"module": "node20",
 		"moduleResolution": "node16",
-		"target": "ES2022",
+		"target": "ES2024",
 		"lib": [
 			"ES2024",
-			"ESNext.Array",
-			"ESNext.Collection",
-			"ESNext.Iterator",
+			"ES2025.Collection",
+			"ES2025.Iterator",
 			"DOM"
 		],
 
@@ -21,7 +20,6 @@
 		"noImplicitOverride": true,
 		"noImplicitReturns": true,
 		"noImplicitThis": true,
-		"strict": true,
 
 		"pretty": true,
 		"declaration": true,
@@ -31,6 +29,7 @@
 		"incremental": true,
 		"newLine": "LF",
 		"skipLibCheck": true,
+		"stableTypeOrdering": true,
 		"types": ["node"]
 	}
 }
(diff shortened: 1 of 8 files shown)


packages/tsconfig-base.json
```diff
@@ -4,12 +4,8 @@
 		"esModuleInterop": true,
 		"module": "node20",
 		"moduleResolution": "node16",
-		"target": "ES2022",
```
NeunEinser, line 7: This means the target is ES2025 the same as lib now?

We are running in node, so I suppose we shouldn't have compatibility issues. Is there a reason we has 2022 as target in the first place?

packages/tsconfig-base.json
```diff
@@ -21,7 +17,6 @@
 		"noImplicitOverride": true,
 		"noImplicitReturns": true,
 		"noImplicitThis": true,
-		"strict": true,
```
NeunEinser, line 24: Any reason this was removed? I tested on your branch and it still runs fine for me with this enabled.

SPGoding: `strict` is the default in TS 6

packages/tsconfig-base.json
SPGoding: > This means the target is ES2025 the same as lib now?

Yes

> We are running in node, so I suppose we shouldn't have compatibility issues. Is there a reason we has 2022 as target in the first place?

Not sure. If we assume the user is on a version of VS Code released this year, they would be on at least Node.js 22.21.1 (source: [vscode-versions](https://github.com/ewanharris/vscode-versions)), which supports almost all of `ES2025` except some RegEx features and `Promise.try` (source: [node.green](https://node.green/)). I guess I could change the `lib` to `ES2024` + `ES2025.Collection` + `ES2025.Iterator` just to be safe (and maybe also bump the minimum VS Code version required by `vscode-extension` to `1.101.0`?)

NeunEinser: Yeah that sounds good to me -- Confirmed 1.101.0 has the same compatibility with node 22.15.

NeunEinser: We should also update `engines` in pacakge.json and `@types/node` to 22.15.

Currently, discord-bot overrides `engines` to 22.12 (from the normal 22.14), so it would also be worth checking what the server running the discord bot currently is running with (though I don't think it matters, 22.12 is actually the first version supporting the exact same compatibility as 22.15/22.21)

NeunEinser (approved):

PR #2016, opened 2026-05-10
👷 Use the workflow actor as the commit author in release step

SPGoding:


scripts/release.ts
```diff
@@ -283,16 +305,28 @@ async function main(): Promise<void> {
...
+			isDryRun,
+			'git',
+			[
+				'remote',
+				'set-url',
+				'origin',
+				`https://${botName}:${githubToken}@github.com/SpyglassMC/Spyglass.git`,
+			],
+			RepoRoot,
+			{ redactArgs: true },
```
SPGoding, line 321: We were relying on GitHub Action to redact our token in the logs but we should probably just not log that in the first place, so added `redactArgs`

PR #2017, opened 2026-05-10
♻️ Simplify Externals

### LINKED ISSUE #1913: Simplify some Externals fields with standard web APIs
[`Externals`](https://github.com/SpyglassMC/Spyglass/blob/main/packages/core/src/common/externals/index.ts) was created to abstract away differences in browser and Node.js environments. With many web APIs now supported natively by Node.js, we can remove some of them and replace their usage with standard web APIs instead.

* `archive.gzip()` and `archive.gunzip()`: replace with [`CompressionStream`](https://developer.mozilla.org/en-US/docs/Web/API/CompressionStream) and [`DecompressionStream`](https://developer.mozilla.org/en-US/docs/Web/API/DecompressionStream), which have been on the global object in Node.js since v18.0.0. We can remove the dependency on `pako` on web as well.
* `crypto.getSha1()`: replace with [`Crypto`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto), which has been supported by Node.js since v15.0.0.
* `event.EventEmitter`: replace with [`EventTarget`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget), which has been supported by Node.js since v15.0.0.
* `web.fetch()`: remove. This value is already the same on both browser and Node.js. We should just inline it with `fetch` from the global object wherever it's used. `Externals` wasn't really created for mocking/dependency injection purposes and it's pretty simple to mock `globalThis.fetch` in tests if needed anyways.

SPGoding:
This PR resolves https://github.com/SpyglassMC/Spyglass/issues/1913.

TODO:
- [x] Probably should add some tests for the replacements

### DIFF
--- a/packages/core/src/common/externals/BrowserExternals.ts
+++ b/packages/core/src/common/externals/BrowserExternals.ts
@@ -154,7 +154,6 @@ export const BrowserExternals: Externals = {
 	event: { EventEmitter: BrowserEventEmitter },
 	fs: new BrowserFileSystem(),
 	web: {
-		fetch,
 		getCache: () => window.caches.open('spyglassmc'),
 	},
 }
--- a/packages/core/src/common/externals/BrowserExternals.ts
+++ b/packages/core/src/common/externals/BrowserExternals.ts
@@ -1,51 +1,6 @@
 import { decode as arrayBufferFromBase64, encode as arrayBufferToBase64 } from 'base64-arraybuffer'
 import { fileUtil } from '../../service/fileUtil.js'
-import type {
-	ExternalEventEmitter,
-	ExternalFileSystem,
-	Externals,
-	ExternalStats,
-	FsLocation,
-} from './index.js'
-
-type Listener = (...args: unknown[]) => unknown
-export class BrowserEventEmitter implements ExternalEventEmitter {
-	readonly #listeners = new Map<string, { all: Set<Listener>; once: Set<Listener> }>()
-
-	emit(eventName: string, ...args: unknown[]): boolean {
-		const listeners = this.#listeners.get(eventName)
-		if (!listeners?.all?.size) {
-			return false
-		}
-		for (const listener of listeners.all) {
-			listener(...args)
-			if (listeners.once.has(listener)) {
-				listeners.all.delete(listener)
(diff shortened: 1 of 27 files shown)


MulverineX (approved):

NeunEinser:
Tests would be nice, lgtm otherwise!

packages/core/src/common/externals/BrowserExternals.ts
```diff
@@ -136,21 +90,6 @@ export const BrowserExternals: Externals = {
...
 		},
-		async gunzip(buffer) {
-			return pako.inflate(buffer) as Uint8Array<ArrayBuffer>
-		},
-		async gzip(buffer) {
-			return pako.gzip(buffer) as Uint8Array<ArrayBuffer>
-		},
-	},
-	crypto: {
-		async getSha1(data) {
```
misode, line 147: I know this is spacebar heating, but I actually overwrote this function as an optimization: https://github.com/misode/misode.github.io/pull/614/changes/46ed105c348c63306093cccf0a7d6c96777faa51

misode (approved):
Nice

packages/core/src/common/externals/BrowserExternals.ts
SPGoding: Interesting, you would think web crypto is implemented with native code which should be faster than JavaScript code even if the algorithm itself is much slower, but that is not the case according to some very non-scientific benchmarking on Firefox:

image

It is the case with Chromium though:

Image

packages/core/test/common/util.spec.ts
```diff
@@ -55,6 +59,84 @@ describe('common util', () => {
...
+					Reflect.defineProperty(Uint8Array.prototype, 'toHex', toHexDescriptor)
+				} else {
+					Reflect.deleteProperty(Uint8Array.prototype, 'toHex')
+				}
+			}
+		})
+	})
+
+	describe('getSha1()', () => {
+		const inputString = 'mapo tofu'
```
NeunEinser, line 116: Looked it up, sounds tasty! :D

NeunEinser (approved):

PR #2018, opened 2026-05-10
✨ Fetcher improvements

SPGoding:
This in preparation for https://github.com/SpyglassMC/Spyglass/issues/1938.

* `HttpCache` / `fetcher`: Adds support for caching based on `last-modified` / `if-modified-since` headers in addition to the existing `etag` / `if-none-match` headers, as the Mojang API uses the time based ones instead of `etag`
* `HttpCache`: Adds support for caching partial requests with the `Range` header, which will be used for extracting `version.json` from client Jar files without downloading the whole 30MB+ thing
* `HttpCache`: Switches to using an index file with a content-addressable storage instead of having the file names themselves as indexing keys in base64url encoding. This was mainly changed due to concerns regarding how to encode both the request URI and the `Range` header in the file name without making the file paths too long (which could cause problems on Windows). Also makes it easier to implement cache eviction in the future if we ever need to as we wouldn't need to walk the full file tree to find out what's cached.
* `fetcher`: Adds retries for network-level problems

### DIFF
--- a/packages/core/src/common/externals/NodeJsExternals.ts
+++ b/packages/core/src/common/externals/NodeJsExternals.ts
@@ -11,7 +11,7 @@ import { pipeline } from 'node:stream/promises'
 import type streamWeb from 'node:stream/web'
 import url from 'node:url'
 import { promisify } from 'node:util'
-import type { DecompressedFile, RootUriString } from '../../index.js'
+import { isObject, type DecompressedFile, type RootUriString } from '../../index.js'
 import { Logger } from '../Logger.js'
 import type { Uri } from '../util.js'
 import type { Externals, FsLocation } from './index.js'
@@ -136,20 +136,20 @@ interface CacheIndex {
 }
 namespace CacheIndex {
 	export function assert(val: unknown): asserts val is CacheIndex {
-		if (!(!!val && typeof val === 'object')) {
+		if (!isObject(val)) {
 			throw new Error('Expected an object')
 		}
-		if (!('index' in val && !!val.index && typeof val.index === 'object')) {
+		if (!('index' in val && isObject(val.index))) {
 			throw new Error("Expected 'index' to exist as an object")
 		}
 		if (
 			!(Object.values(val.index).every((i) =>
-				!!i && typeof i === 'object'
+				isObject(i)
 				&& Object.values(i).every((v) =>
-					!!v && typeof v === 'object'
+					isObject(v)
 					&& 'status' in v && typeof v.status === 'number'
 					&& 'statusText' in v && typeof v.statusText === 'string'
-					&& 'headers' in v && !!v.headers && typeof v.headers === 'object'
+					&& 'headers' in v && isObject(v.headers)
 					&& Object.values(v.headers).every((s) => typeof s === 'string')
 					&& 'sha1' in v && typeof v.sha1 === 'string' && /^[0-9a-f]{40}$/.test(v.sha1)
 					&& 'cacheTime' in v && typeof v.cacheTime === 'number'
--- a/packages/core/src/common/externals/NodeJsExternals.ts
+++ b/packages/core/src/common/externals/NodeJsExternals.ts
@@ -2,7 +2,7 @@ import decompress from 'decompress'
 import { Buffer } from 'node:buffer'
 import cp from 'node:child_process'
 import { createHash, randomUUID } from 'node:crypto'
-import fs from 'node:fs'
+import type fs from 'node:fs'
 import fsp from 'node:fs/promises'
 import os from 'node:os'
 import process from 'node:process'
@@ -93,7 +93,7 @@ export function getNodeJsExternals(
 			},
 			web: {
 				getCache: async () => {
-					return new HttpCache(cacheRoot, logger)
+					return new HttpCache(cacheRoot, logger, nodeFsp)
 				},
 			},
 		} satisfies Externals,
@@ -167,20 +167,25 @@ namespace CacheIndex {
  */
 class HttpCache implements Cache {
 	readonly #cacheRoot: RootUriString | undefined
+	readonly #httpRoot: RootUriString | undefined
 	readonly #indexUri: string | undefined
 	readonly #objectsRoot: RootUriString | undefined
 	readonly #tempRoot: RootUriString | undefined
 	readonly #logger: Logger
+	readonly #fsp: typeof fsp
 	#index: CacheIndex | undefined
+	#loadIndexPromise: Promise<void> | undefined
 
-	constructor(cacheRoot: RootUriString | undefined, logger: Logger) {
+	constructor(cacheRoot: RootUriString | undefined, logger: Logger, nodeFsp: typeof fsp) {
 		if (cacheRoot) {
-			this.#cacheRoot = `${cacheRoot}http/`
-			this.#indexUri = `${this.#cacheRoot}index.json`
-			this.#objectsRoot = `${this.#cacheRoot}objects/`
-			this.#tempRoot = `${this.#cacheRoot}temp/`
+			this.#cacheRoot = cacheRoot
+			this.#httpRoot = `${this.#cacheRoot}http/`
+			this.#indexUri = `${this.#httpRoot}index.json`
+			this.#objectsRoot = `${this.#httpRoot}objects/`
+			this.#tempRoot = `${this.#httpRoot}temp/`
 		}
 		this.#logger = logger
+		this.#fsp = nodeFsp
 	}
 
 	async match(
@@ -200,7 +205,7 @@ class HttpCache implements Cache {
 		}
 
 		try {
-			const objectFileHandle = await fsp.open(
+			const objectFileHandle = await this.#fsp.open(
 				this.#getObjectUri(this.#objectsRoot, indexEntry.sha1),
 			)
 			const bodyStream = objectFileHandle.createReadStream({ autoClose: true })
@@ -214,6 +219,7 @@ class HttpCache implements Cache {
 			)
 		} catch (e) {
 			if ((e as NodeJS.ErrnoException)?.code === 'ENOENT') {
+				this.#logger.warn(`Object file for ${JSON.stringify(indexEntry)} does not exist`, e)
 				delete index.index[requestUri]?.[requestRange ?? '']
 				if (Object.values(index.index[requestUri]!).length === 0) {
 					delete index.index[requestUri]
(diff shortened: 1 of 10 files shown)


packages/core/src/common/util.ts
```diff
@@ -345,6 +345,10 @@ export function normalizeUri(uri: string): string {
 	return obj.toString()
 }
 
+export async function sleep(delayMs: number): Promise<void> {
```
NeunEinser: (nit) `async` shouldn't be needed here as you don't `await` anything

packages/core/src/service/fetcher.ts
```diff
@@ -42,10 +50,52 @@ export async function fetchWithCache(
...
+	for (let i = 0; i < FETCH_RETRY_MAX_ATTEMPTS; i++) {
+		const isLastAttempt = i === FETCH_RETRY_MAX_ATTEMPTS - 1
+		try {
+			lastResult = await web.fetch(request.clone(), {
+				signal: AbortSignal.any([
+					totalSignal,
+					AbortSignal.timeout(FETCH_PER_TRY_TIMEOUT_MS),
+				]),
+			})
+			if (lastResult.status < 500) {
```
NeunEinser, line 105: Would it make sense here to also check if the response is not stale?

SPGoding: Hmm I don't think so. Stale response comes from the outer `fetchWithCache()` function after all the retries have failed. `fetchWithRetries()` doesn't know about cache or anything

NeunEinser: Oh right fsr I thought it was inside fetchWithCache

NeunEinser (approved):

packages/core/src/common/util.ts
SPGoding: Fixed in https://github.com/SpyglassMC/Spyglass/pull/2018/changes/903ff176279cff19982695147629cd536bb84282 ([diff](https://github.com/SpyglassMC/Spyglass/compare/d7d249e1afc5c2771913e5b3442d3fe77178f72e..903ff176279cff19982695147629cd536bb84282))

packages/core/src/common/externals/NodeJsExternals.ts
```diff
@@ -113,42 +121,111 @@ function toPath(path: FsLocation): string {
...
+			} | undefined
+		} | undefined
+	}
+}
+namespace CacheIndex {
+	export function assert(val: unknown): asserts val is CacheIndex {
+		if (!(!!val && typeof val === 'object')) {
+			throw new Error('Expected an object')
+		}
+		if (!('index' in val && !!val.index && typeof val.index === 'object')) {
```
NeunEinser: Had to look up why `!!val.index` is needed. I think the only way this would be falsey is if val.index is `null`. Empty object is truly, `typeof undefined` would be `"undefined"`. Would maybe be a bit clearer if it just spelled out `val.index !== null`

packages/core/src/common/externals/NodeJsExternals.ts
```diff
@@ -113,42 +121,111 @@ function toPath(path: FsLocation): string {
...
+	readonly #indexUri: string | undefined
+	readonly #objectsRoot: RootUriString | undefined
+	readonly #tempRoot: RootUriString | undefined
+	readonly #logger: Logger
+	readonly #fsp: typeof fsp
+	#index: CacheIndex | undefined
+	#loadIndexPromise: Promise<void> | undefined
 
-	constructor(cacheRoot: RootUriString | undefined) {
+	constructor(cacheRoot: RootUriString | undefined, logger: Logger, nodeFsp: typeof fsp) {
```
NeunEinser, line 179: Is the `nodeFsp` parameter just for mocking? I have never seen a namespace type like that but if it works it works? 🤷

packages/core/src/common/externals/NodeJsExternals.ts
```diff
@@ -157,28 +234,164 @@ class HttpCache implements Cache {
...
+		const requestRange = request instanceof Request ? request.headers.get('range') : undefined
+
+		const bodySha1 = await this.#saveResponseBody(
+			response.body as streamWeb.ReadableStream,
+			this.#tempRoot,
+			this.#objectsRoot,
+		)
+
+		const index = await this.#loadIndex(this.#indexUri)
+		index.index[requestUri] ??= Object.create(null)
```
NeunEinser: What's the difference between `Object.create(null)` and the simple `{}` literal (if any)?

packages/core/src/common/externals/NodeJsExternals.ts
```diff
@@ -157,28 +234,164 @@ class HttpCache implements Cache {
...
+		await this.#fsp.mkdir(new URL(tempRoot), { recursive: true, mode: 0o755 })
+		const tempHandle = await this.#fsp.open(tempUri, 'w', 0o644)
+		const writeStream = tempHandle.createWriteStream({ autoClose: true })
+		bodyStream.pipe(bodyHash)
+		await pipeline(bodyStream, writeStream)
+
+		// Move the temporary file to the SHA-1 Content-addressable objects store
+		const bodySha1 = bodyHash.digest('hex')
+		const objectUri = this.#getObjectUri(objectsRoot, bodySha1)
+		await this.#fsp.mkdir(new URL('.', objectUri), { recursive: true, mode: 0o755 })
```
NeunEinser: Is the `'.'` parameter needed, or could you also just do `URL(objectUri)` without a base?

packages/core/src/common/externals/NodeJsExternals.ts
```diff
@@ -157,28 +234,164 @@ class HttpCache implements Cache {
...
+					this.#logger.warn('[HttpCache] Corrupted cache index', e)
+				}
+				this.#index = { index: {} }
+			}
+		})())
+		return this.#index!
+	}
+
+	async #saveIndex(indexUri: string, tempRoot: RootUriString): Promise<void> {
+		try {
```
NeunEinser, line 321: `Dev.assertDefined(this.#index)`?

packages/core/src/common/externals/NodeJsExternals.ts
```diff
@@ -157,28 +234,164 @@ class HttpCache implements Cache {
...
+	}
+
+	/**
+	 * Remove cache files used by older versions of Spyglass.
+	 * - until v2026.5.8+8c7f6e, `${cacheRoot}downloader/`
+	 * - until v2026.5.16+9c4fa2, `${cacheRoot}http/${base64UrlEncoded}.${'bin' | 'etag'}`
+	 */
+	async #cleanUpLegacyCache(): Promise<void> {
+		// Stupid hardcoded path check to make sure we don't accidentally delete some random files somehow
+		if (!(this.#cacheRoot && this.#cacheRoot.endsWith('/spyglassmc-nodejs/') && this.#httpRoot)) {
```
NeunEinser: I don't really see why this would be needed. I guess we could be stupid and accidentally pass the cwd?

I dunno, there could be some change, like the cacheroot not ending with a `/` anymore, that could break this and I don't really see a way this could point to a random directory and delete a random file (if we trust our `env-paths` dep)

packages/core/src/common/externals/NodeJsExternals.ts
```diff
@@ -157,28 +234,164 @@ class HttpCache implements Cache {
...
+			} catch (e) {
+				if ((e as NodeJS.ErrnoException).code === 'ENOENT') {
+					// Triger legacy cache clean up if no index file for the new cache scheme exists.
+					this.#logger.info('[HttpCache] No cache index found; cleaning up legacy cache files')
+					await this.#cleanUpLegacyCache()
+				} else {
+					this.#logger.warn('[HttpCache] Corrupted cache index', e)
+				}
+				this.#index = { index: {} }
+			}
```
NeunEinser, line 315: Is it intended to mute all exceptions here, or should there be a `throw` after the closing `}` from the if?

packages/core/src/common/externals/NodeJsExternals.ts
```diff
@@ -157,28 +234,164 @@ class HttpCache implements Cache {
...
+		}
+
+		const requestUri = request instanceof Request ? request.url : request.toString()
+		const requestRange = request instanceof Request ? request.headers.get('range') : undefined
+
+		const bodySha1 = await this.#saveResponseBody(
+			response.body as streamWeb.ReadableStream,
+			this.#tempRoot,
+			this.#objectsRoot,
+		)
```
NeunEinser: Would it make sense to save the response body at the end (and get the sha here separately), to avoid having saved a file, and after the saving of the index fails for some reason?

The index containing a file that does not exist on disk is already handled gracefully

packages/core/src/common/externals/NodeJsExternals.ts
SPGoding: `new URL('.', 'file:///root/file')` returns `new URL('file:///root/')` (the parent directory of the file) so yes it's needed.

packages/core/src/common/externals/NodeJsExternals.ts
SPGoding: Yes

packages/core/src/common/externals/NodeJsExternals.ts
SPGoding: `Object.create(null)` has a `null` prototype so it doesn't have all the keys normal objects have inherited from `Object.prototype` (e.g. `Object.prototype.hasOwnProperty`). I just do that for objects used as a hashmap. Probably doesn't matter for our case since this will end up becoming a normal object with `Object.prototype` after JSON serialization anyways. I can remove it.

packages/core/src/common/externals/NodeJsExternals.ts
SPGoding: Yes it's intentional.

packages/core/src/common/externals/NodeJsExternals.ts
SPGoding: I don't want to buffer the entire response body in memory so the caculating the sha1 and saving it to the file is done in one go through a stream tee, so not sure how to calculate sha1 first without writing it. I guess I could only write it to the temporary file in the first step, and after the index is written I rename it to the final object location?

packages/core/src/common/externals/NodeJsExternals.ts
SPGoding: I've seen so many horror stories of software deleting the user's files (e.g. [Steam](https://github.com/ValveSoftware/steam-for-linux/issues/3671), [Minecraft Dungeons Launcher](https://bugs.mojang.com/browse/MCL/issues/MCL-14107)) so added this check just in case. Might not be strictly necessary since we are only deleting specific directories named `downloader` and files named `http/*.bin` and `http/*.etag` within the cache root.

`cacheRoot` has a type level constraint (`RootUriString`) to enforce a trailing slash so I don't think that would change. The only case this check would miss is if third-party uses of Spyglass core (maybe Misode generator?) use a different cache root.

packages/core/src/common/externals/NodeJsExternals.ts
NeunEinser: Ah interesting. I don't mind it.

packages/core/src/common/externals/NodeJsExternals.ts
NeunEinser: That would work, yes. I don't think it's a big deal but this does sound like a simple-enough solution.

packages/core/src/common/externals/NodeJsExternals.ts
SPGoding: Is this better: https://github.com/SpyglassMC/Spyglass/pull/2018/changes/93a37205597b1e4c462e2329e9834a5d0c4047de

packages/core/src/common/externals/NodeJsExternals.ts
SPGoding: https://github.com/SpyglassMC/Spyglass/pull/2018/commits/1c52f78e091b18e44f9430b06999acb2f089c64e

packages/core/src/common/externals/NodeJsExternals.ts
SPGoding: Removed: https://github.com/SpyglassMC/Spyglass/pull/2018/commits/0ab8aaeaf223bbf015bf1a84616f651e77c9bcb2

packages/core/src/common/externals/NodeJsExternals.ts
SPGoding: https://github.com/SpyglassMC/Spyglass/pull/2018/commits/dcc4fb97f15bc47b5be42f3db12e434aa4f53029

packages/core/src/common/externals/NodeJsExternals.ts
NeunEinser: Yeah tbf this was more of a nit anyways.

NeunEinser (approved):
But I hardly know 'er!

PR #2019, opened 2026-05-10
✨ Implement robust versions meta-resource handling

SPGoding:
This PR builds on top of:

- https://github.com/SpyglassMC/Spyglass/pull/2017
- https://github.com/SpyglassMC/Spyglass/pull/2018

It handles the `versions` aspect of https://github.com/SpyglassMC/Spyglass/issues/1938.

TODO:

- [x] On Mojang API path, try to use range request to download `version.json` files from client jars to obtain `data_major` / `data_minor` / `resource_major` / `resource_minor`. Because this uses range request it should be relatively cheap without having to transmit the whole ~30MB of client jar, but need testing to confirm
- [x] Current `zip.js` usage does not cache the client jar range requests. Maybe add range request caching to our fetcher and pass it to `HttpRangeRequest`?
- [x] Add tests


packages/java-edition/scripts/fetch_bundled_meta_resources.ts
```diff
@@ -0,0 +1,32 @@
...
+	const githubApiResult = await (await fetch(GITHUB_API_URI)).json()
+	McmetaVersions.assert(githubApiResult)
+	await fsp.writeFile(versionsOutputPath, JSON.stringify(githubApiResult) + '\n')
+}
+
+try {
+	await main()
+} catch (e) {
+	console.error(e)
+	process.exit([42, 67, 69, 91][Math.floor(Math.random() * 4)])
```
NeunEinser, line 31: 🤔 😁

packages/java-edition/src/dependency/mcmeta/types.ts
```diff
@@ -0,0 +1,201 @@
...
+
+export type McmetaVersions = [McmetaVersion, ...McmetaVersion[]]
+export namespace McmetaVersions {
+	export function assert(val: unknown): asserts val is McmetaVersions {
+		if (!Array.isArray(val) || val.length === 0) {
+			throw new Error('Expected non-empty array')
+		}
+		const badVersion = val.find((v) => !McmetaVersion.is(v))
+		if (badVersion) {
+			throw new Error(`Found malformed McmetaVersion: ${JSON.stringify(badVersion)}`)
```
NeunEinser: (nit) What would you think about having some kind of function in `McmetaVersion` which gives an error code? That way we could throw something like “Found malformed McmetaVersion: No key `resource_pack_version` in `{ "name": "Fetchr 9.1-beta69" }`”. `is` is probably still nice and could just check for that other function returning `undefined`.

packages/java-edition/src/dependency/mcmeta/versions.ts
```diff
@@ -0,0 +1,271 @@
...
+		result = await import('./versions.json', { with: { type: 'json' } })
+		McmetaVersions.assert(result)
+		return result
+	} catch (e) {
+		logger.warn('Load bundled McmetaVersions failure', e, result)
+	}
+	return undefined
+}
+
+async function fetchMojangApi<T>(
```
NeunEinser: Is there anything destinctly specific to mojang API in this function (besides the log warning)? Could maybe just be a `fetchWithTypeAssertion` or so in fetcher.

packages/java-edition/src/dependency/mcmeta/versions.ts
```diff
@@ -0,0 +1,271 @@
...
+	if (!mojangVersionManifest) {
+		return versions
+	}
+
+	const existingMcmetaVersions = new Set(versions.map((v) => v.id))
+	const newMojangVersionManifestEntries: MojangVersionManifestEntry[] = []
+	for (const mojangVersionManifestEntry of mojangVersionManifest.versions) {
+		if (existingMcmetaVersions.has(mojangVersionManifestEntry.id)) {
+			break
+		}
```
NeunEinser, line 159: I wonder if this could be done more efficiently, e.g. by looking at the version id of the topmost `versions` entry and then looking for that version in the mojang manifest and assuming everything after that is already covered.

This would rely on both our API and MojangAPI to always return a correctly sorted list (which they really should), but maybe it's not a good idea to rely on this?

One issue that also I see is that our version API starts with 1.14, and Mojang's manifest obviously contains way more (all the way up to some classic versions I think; it's 887 vs 430 entries), so if you keep this you should at least also filter versions below the minimum version in the versions array (e.g. by releaseTime?)

packages/java-edition/src/dependency/mcmeta/versions.ts
```diff
@@ -0,0 +1,271 @@
...
+			}
+		}
+		if (!versionJsonEntry) {
+			logger.info(`Client Jar from '${clientJarUri}' does not have 'version.json'`)
+			return undefined
+		}
+
+		const versionJsonText = await versionJsonEntry.getData(new TextWriter())
+		versionJson = JSON.parse(versionJsonText)
+		core.Dev.assertTrue(MojangVersionJson.is(versionJson), 'Invalid MojangVersionJson')
```
NeunEinser: Should this not be an actual assertion that checks this file is valid in production code, like the other assertions?

packages/java-edition/src/dependency/mcmeta/versions.ts
```diff
@@ -0,0 +1,271 @@
...
+
+async function fetchMojangVersionJsonFromClientJar(
+	{ logger, clientJarUri }: {
+		logger: core.Logger
+		clientJarUri: string
+	},
+): Promise<MojangVersionJson | undefined> {
+	let versionJson: unknown
+	try {
+		const zipReader = new ZipReader(new HttpRangeReader(clientJarUri))
```
NeunEinser: no `fetchWithCache`?

I suppose instead of caching the entire jar, it could also make sense to cache the resulting hydrated versions.json, at least for versions for which for which the `latestMcmetaVersion` fallback didn't need to get used (which could be a simple flag and filtered), and then use that cache instead of the bundled version.json when it exists and is valid.

This might cause some issues with the previously discussed assertion that all versions are in the correct order and complete though.

packages/java-edition/src/dependency/mcmeta/versions.ts
SPGoding: I think this is already sorta optimized due to the use of `break` in the loop. It assumes Mojang API is sorted correctly, so once we find a version in Mojang API that is already in `mcmetaVersions` it `break`s the loop and stops checking (note that both `mcmetaVersions` and `mojangVersionManifestEntry[]` are sorted in reverse chronological order, so the first entry in the latest.). I could optimize it more by checking against the first entry in `mcmetaVersions` instead of adding all of `mcmetaVersions` to a `Set`, but that's a pretty cheap operation anyways with only 88 versions currently existing in `mcmetaVersions` so I don't think that's really needed.

packages/java-edition/src/dependency/mcmeta/versions.ts
SPGoding: See TODO :)

This does not require fetch/caching the whole jar as `zip.js` uses HTTP range requests under the hood to download just the metadata and `version.json` from the jar. We need to update the fetchWithCache() to support caching range requests.

As for caching hydrated `versions.json` I don't see a good place to do that and it'll probably be unnecessary since all the artifacts leading to that hydration would be cached instead

packages/java-edition/src/dependency/mcmeta/versions.ts
SPGoding: Maybe. I was a bit willy nilly with `.is()` vs `.assert()` since those are quite annoying to write

packages/java-edition/src/dependency/mcmeta/versions.ts
NeunEinser: 88? Is it not including all snapshots, etc?

I misread the `break` as `continue` 🙃

packages/java-edition/src/dependency/mcmeta/versions.ts
NeunEinser: Ohh nice, TIL range requests exist :)

Yeah I agree, if we can just cache the single file that is returned from this request, this is enough.

packages/java-edition/src/dependency/mcmeta/versions.ts
SPGoding: It's 430, no idea why I thought it was 88. Still pretty cheap to add to a set though imo. I think I was trying to avoid failure case where the first version in mcmeta is somehow not in Mojang API and we'd loop through the entire Mojang API in that case. But maybe adding e.g. the first 100 entries from mcmeta to the Set would be enough.

SPGoding:
Force-pushed to rebase on top of latest https://github.com/SpyglassMC/Spyglass/pull/2017

SPGoding:
Force-pushed to rebase on top of latest https://github.com/SpyglassMC/Spyglass/pull/2018. No changes; will get to this PR some time later :)

packages/java-edition/src/dependency/mcmeta/types.ts
SPGoding: Something like this? https://github.com/SpyglassMC/Spyglass/pull/2019/changes/a0366663b4dd165dcc4eb796189d9b67c8d1ea4d

packages/java-edition/src/dependency/mcmeta/versions.ts
SPGoding: https://github.com/SpyglassMC/Spyglass/pull/2019/commits/81e9644e653953c4788c3c86e4a932666374c26c

packages/java-edition/src/dependency/mcmeta/versions.ts
SPGoding: Mapping through 430 entries is basically free, so I'll leave it as-is. Added some comments in https://github.com/SpyglassMC/Spyglass/pull/2019/changes/d0816856fc6d059c36e3fe44e2e62605a19699ca

packages/java-edition/src/dependency/mcmeta/versions.ts
SPGoding: Added `fetchWithCache`: https://github.com/SpyglassMC/Spyglass/pull/2019/changes/43dfdfb5283c4ed6b4de49336b2db2ecf42a1e0b

packages/java-edition/src/dependency/mcmeta/versions.ts
SPGoding: https://github.com/SpyglassMC/Spyglass/pull/2019/commits/96be9ccb555cd713c056a4c8b9bc24119de77784

packages/java-edition/src/dependency/mcmeta/types.ts
NeunEinser: Yes!

packages/java-edition/src/dependency/mcmeta/versions.ts
```diff
@@ -0,0 +1,275 @@
...
+	MojangVersionJson,
+	MojangVersionManifest,
+} from './types.js'
+
+const SPYGLASSMC_API_URI = 'https://api.spyglassmc.com/mcje/versions'
+const GITHUB_API_URI =
+	'https://raw.githubusercontent.com/misode/mcmeta/refs/heads/summary/versions/data.min.json'
+const MOJANG_VERSION_MANIFEST_URI =
+	'https://piston-meta.mojang.com/mc/game/version_manifest_v2.json'
+const MAX_CLIENT_JAR_FETCHES = 5
```
NeunEinser, line 17: 5 seems quite limiting. I suppose you are afraid of these calls being potentially slow and tank startup performance, especially if we didn't update the extension in a while or sb. has an older version?

Are these assumptions correct?
- The order in which Mojang versions are attempted to be fetched is the latest one first, then the next snapshot, etc?
- The way the cache works, the resulting full manifest is never stored, instead the individual requests are cached (so the Mojang version_manifest would be stored, and each individual client call)

Potential optimization here:
Only fetch the latest version of each missing version cycle. With the new version format, it is always possible to know the target release version of any snapshot. We currently only care for the release version in Spyglass.

So we really only need the “latest” version of a cycle, and if we get some pack_format in between, we can fallback to the next next higher version we do know

I would like this more to reduce calls. The top 5 of the list are probably not the most useful.

I am fine with doing putting this off though, maybe make an issue and think about it more, and leaving this as-is for the initial merge.

NeunEinser (approved):

MulverineX (approved):

packages/java-edition/src/dependency/mcmeta/versions.ts
SPGoding: https://github.com/SpyglassMC/Spyglass/issues/2068

PR #2020, opened 2026-05-12
⬆️ TypeScript 7

SPGoding:
Update tsconfig project references to reference dependencies at `src` and `test` level instead of at package level. See also [microsoft/typescript-go#3795](https://redirect.github.com/microsoft/typescript-go/issues/3795)

The build now passes with `tsgo`, so switched build scripts to use `tsgo` instead.

### DIFF
--- a/.packages.json
+++ b/.packages.json
@@ -1,4 +1,15 @@
 {
+	"benchmarks": {
+		"dependencies": [
+			"core",
+			"java-edition",
+			"json",
+			"mcdoc"
+		],
+		"released": {
+			"private": true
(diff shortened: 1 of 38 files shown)


NeunEinser (approved):
Works on my machine.

I would also be fine with removing the `typescipt` dependency and `build:tsc6` entirely.

SPGoding:
> I would also be fine with removing the typescipt dependency

We still need it for `typescript-eslint` as `tsgo` doesn't have an API for `eslint` yet

PR #2050, opened 2026-07-08
🐛 Fix #2048 no longer allow scheduling functions in the same tick

### LINKED ISSUE #2048: schedule function foo 0
command `schedule function foo 0` is allowed, but actually it would lead into error, since functions cant be scheduled to same tick (26.2)

misode:
Also fixes a related issue in the core float parser when min and max weren't checked when `0`

### DIFF
--- a/packages/core/src/parser/float.ts
+++ b/packages/core/src/parser/float.ts
@@ -88,7 +88,8 @@ export function float(options: Options): Parser<FloatNode> {
 		} else if (!options.pattern.test(raw)) {
 			ctx.err.report(localize('parser.float.illegal', options.pattern), ans)
 		} else if (
-			(options.min && ans.value < options.min) || (options.max && ans.value > options.max)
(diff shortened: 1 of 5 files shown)


packages/core/src/parser/float.ts
```diff
@@ -88,7 +88,8 @@ export function float(options: Options): Parser<FloatNode> {
 		} else if (!options.pattern.test(raw)) {
 			ctx.err.report(localize('parser.float.illegal', options.pattern), ans)
 		} else if (
-			(options.min && ans.value < options.min) || (options.max && ans.value > options.max)
+			(options.min !== undefined && ans.value < options.min)
+			|| (options.max !== undefined && ans.value > options.max)
```
NeunEinser, line 92: Good fix, yeah that would mess it up 😅

NeunEinser (approved):
Nice!

PR #2054, opened 2026-07-11
🐛 Declare @spyglassmc/locales as an explicit workspace dependency

EllaCoat:
Rework following @SPGoding's review: switched from hand-editing 6 `package.json` files to a script that treats `.packages.json` as the single source of truth.

* Renamed `scripts/set_tsconfig_references.ts` → `scripts/sync_package_taxonomy.ts` (npm script entry follows: `set-tsconfig-references` → `sync-package-taxonomy`).
* The script now regenerates each workspace `package.json`'s `dependencies` from `.packages.json`, in addition to the existing tsconfig references. The managed set is built from the actual `@spyglassmc/*` names present in workspace manifests (not the `.packages.json` keys), so external entries and unmanaged names are preserved.
* Pre-flight validation fails fast on unknown dependency names, missing workspace directories, or `name` mismatches on referenced workspaces.
* Ran the script to regenerate 9 workspace `package.json`s and updated `package-lock.json` accordingly (`npm ci --dry-run` passes).
* The new npm script entry uses the explicit `.ts` extension. The old `set-tsconfig-references` entry was extensionless and would fail with `MODULE_NOT_FOUND` on Node 22; this rename fixes that too.

Follow-ups I've kept out of this PR to keep it reviewable — happy to fold any of these in if you'd prefer:

* CI check for taxonomy drift (`npm run sync-package-taxonomy` + `git diff --exit-code -- packages`, or a `--check` flag on the script).
* Sharing the manifest-regeneration logic with `release.ts`'s `setPackageJsons()`.
* Atomic file writes (read/validate/render everything before writing, plus tempfile+rename).
* Fixtures/tests for idempotence, empty deps, typos, and external-entry preservation.

---

_Note: I'm not a native English speaker, so please feel free to point out any awkward phrasing._


SPGoding (commented):
Workspace package dependency is set at release time through [`release.ts`](https://github.com/SpyglassMC/Spyglass/blob/940b1bd658e98ddce503acd35a4eb9c26a7762ce/scripts/release.ts#L123-L126), but it is an oversight on our side that there are package managers depending on them existing at development time.

To ensure `.packages.json` is the single source of truth for package taxonomy, would you mind making a change to [`scripts/set_tsconfig_references.ts`](https://github.com/SpyglassMC/Spyglass/blob/main/scripts/set_tsconfig_references.ts) (maybe rename to `sync_package_taxonomy.ts` or something) to write dependency packages under `packages/*/packages.json` and use that script to update all the package manifests?

EllaCoat:
Thanks for the review — the SoT-based approach is much better than my initial hand-editing, and I've rebased the branch accordingly.

SPGoding (approved):
Thanks!

MulverineX:
wow does this mean that pnpm is supported now for Spyglass development?

EllaCoat:
Yeah, I've been using pnpm on my fork — that's how I ran into the phantom-`locales` failure. Now that workspace deps are declared, the remaining pieces for real pnpm support look pretty small.

MulverineX:
> the remaining pieces for real pnpm support look pretty small.

So there is still stuff that needs to be done before pnpm is fully supported?

EllaCoat:
Based on what I ran into on my fork, here's roughly what we'd still need:

- `workspace:*` protocol on workspace deps (the script currently outputs `"*"`, which npm workspaces treat as a workspace link, but recent pnpm resolves to the registry version even with `link-workspace-packages=true`)
- `pnpm-workspace.yaml` at the root
- `packageManager` field in the root `package.json`
- A committed pnpm lockfile, with CI switched from `npm ci` to `pnpm install --frozen-lockfile`

That should cover the core setup. If this is something the project wants to support upstream, happy to open an issue and put together a PR. Just let me know what makes sense for the current scope.

MulverineX:
Would it be possible to *support* pnpm without moving our CI to it? @EllaCoat

EllaCoat:
Actually, I dug into this a bit more, and it turns out we can keep `npm ci` on CI as-is. The current `"*"` versions work for both package managers as long as pnpm has `linkWorkspacePackages: true`.

The one critical gotcha is that `saveWorkspaceProtocol: false` is basically mandatory in pnpm's config. Without it, `pnpm add` rewrites `"*"` to `"workspace:^"`, which instantly breaks `npm ci`.

I'll try this minimal setup on my fork first to make sure it actually holds up. If it does, happy to open an issue and PR.

MulverineX:
We would appreciate it if you could get it working and do a PR

PR #2062, opened 2026-07-21
✨ Add "Analyze Project" command

### LINKED ISSUE #1375: Add "Analyze Project" command
The language server only parses and binds all files in the project during the `project#ready` stage. It doesn't check or lint them, because that might be a very expensive operation. There should be a command that forces the language server to iterate over all files and check+lint them.

Hardel-DW:
Closes #1375 

Adds a `Spyglass: Analyze Project` command that checks + lints every file in the project and puts everything in the Problems view. Really handy when migrating a pack to a new game version 🙂 ;3

How it works:
- New `Project#analyzeProject()` in core, it awaits READY first, then runs the full pipeline (read → parse → bind → check/lint) on each project file sequentially. 
  
- New `spyglassmc/analyzeProject` request, wired up the same way as `resetProjectCache`. Progress is server-driven (percentage + cancel button), cancellation stops cleanly between two files.

- Added an integration test for `analyzeProject` (memfs + a tiny fake language).

### DIFF
--- a/docs/user/commands.adoc
+++ b/docs/user/commands.adoc
@@ -4,6 +4,12 @@
 
 == Commands
 
+=== `spyglassmc.analyzeProject`
+
+Check and lint every file in the current project and publish the results to the editor's diagnostics panel (e.g. the Problems view in VS Code).
+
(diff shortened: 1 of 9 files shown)


Hardel-DW:
Tested it on Yggdrasil (adds 20 structures, 1200 files) and DnT (7400 files).
CKo4wiV5hx
Code_fi1Hs66Wew

PR #2070, opened 2026-08-10
✨️ 1.21.5 SNBT Changes

### LINKED ISSUE #1771: Support 25w09a SNBT syntax

MulverineX:
Related to #1771 , team hasn't decided whether this will get to close that issue yet


packages/core/src/parser/float.ts
```diff
@@ -56,12 +56,13 @@ export function float(options: Options): Parser<FloatNode> {
 		if (src.peek() === '-' || src.peek() === '+') {
 			src.skip()
 		}
-		while (src.canRead() && Source.isDigit(src.peek())) {
+		// Digit reader accepts `_` between digits as a separator (ES2021-style numeric separators).
+		while (src.canRead() && (Source.isDigit(src.peek()) || src.peek() === '_')) {
```
misode, line 59: This is not valid for all `float` parser calls. This parser is also used in the json and mcdoc parsers. We could make an argument to also allow this in mcdoc, but then the spec would also have to be updated.

It is also not valid in versions before 1.21.5, so when this parser is called from the nbt parser a version check needs to be made and passed as a config.

This likely applies to all of the other modifications in this PR because I don't see any version checks.

misode (changes requested):

packages/core/src/parser/float.ts
MulverineX: image

MulverineX: image

MulverineX: image

image

MulverineX: The other new syntax samples do error but the error reporting is a little bugged/can be improved, I'll work on that

MulverineX: image
image
image
image

packages/core/src/parser/float.ts
```diff
@@ -56,12 +56,13 @@ export function float(options: Options): Parser<FloatNode> {
 		if (src.peek() === '-' || src.peek() === '+') {
 			src.skip()
 		}
-		while (src.canRead() && Source.isDigit(src.peek())) {
+		// Digit reader accepts `_` between digits as a separator (ES2021-style numeric separators).
```
SPGoding: This comment explains nothing. "ES2021-style" is also completely irrelevant

packages/java-edition/src/mcfunction/checker/index.ts
```diff
@@ -357,6 +360,18 @@ function getTypesFromEntity(
...
+export const uuid: core.Checker<UuidNode> = (node) => {
+	if (node.bits[0] === 0n && node.bits[1] === 0n) {
+		return
+	}
+	const parts: number[] = []
+	for (const bits of node.bits) {
+		parts.push(Number(BigInt.asIntN(32, bits >> 32n)))
+		parts.push(Number(BigInt.asIntN(32, bits & 0xFFFFFFFFn)))
+	}
+	node.hover = `\`\`\`mcdoc\n[I; ${parts.join(', ')}]\n\`\`\``
```
SPGoding: Why `mcdoc` instead of `snbt` (do we not have a TextMate grammar for snbt)

packages/java-edition/src/mcfunction/checker/index.ts
```diff
@@ -357,6 +360,18 @@ function getTypesFromEntity(
 	return undefined
 }
 
+export const uuid: core.Checker<UuidNode> = (node) => {
```
SPGoding: Why is this a checker? Can't this be done in the parser (synchronous with no symbol access needed)?

packages/nbt/src/colorizer/index.ts
```diff
@@ -1,15 +1,64 @@
...
+		),
+		ColorToken.create(
+			core.Range.create(node.prefixRange.start + 4, node.prefixRange.end),
+			'operator',
+		),
+		ColorToken.create(node.suffixRange, 'operator'),
+	]
+}
+
+const compound: core.Colorizer<NbtCompoundNode> = (node, ctx) => {
```
SPGoding, line 43: Is the only purpose of this to create `operator` tokens for the colon and the comma? If so I do not like it as (1) they are not operators so the token is wrong, and (2) our semantic coloring is already coloring too much stuff; syntactical level stuff should be left to the much faster TextMate grammar on VS Code / tree-sitter on other editors.

packages/nbt/src/colorizer/index.ts
```diff
@@ -1,15 +1,64 @@
...
+	return [
+		ColorToken.create(
+			core.Range.create(node.prefixRange.start, node.prefixRange.start + 4),
+			'function',
+		),
+		ColorToken.create(
+			core.Range.create(node.prefixRange.start + 4, node.prefixRange.end),
+			'operator',
+		),
+		ColorToken.create(node.suffixRange, 'operator'),
```
SPGoding, line 39: Same complaint for the `operator` token here

packages/nbt/src/parser/primitive.ts
```diff
@@ -119,8 +161,38 @@ export const primitive: core.InfallibleParser<NbtPrimitiveNode> = (
 		src,
 		ctx,
 	)
+	const isNewSyntax = newSyntax(ctx)
+	let underscoreNotified = false
```
SPGoding: Is there a reason we do this?

packages/nbt/src/parser/primitive.ts
```diff
@@ -119,8 +161,38 @@ export const primitive: core.InfallibleParser<NbtPrimitiveNode> = (
...
+					core.ErrorSeverity.Error,
+				)
+			}
+			// Underscore digit separators are only valid in new syntax (1.21.5+).
+			if (e.group !== Group.Boolean && !isNewSyntax && unquotedResult.value.includes('_')) {
+				if (!underscoreNotified) {
+					ctx.err.report(
+						localize('nbt.parser.number.underscore-not-supported'),
+						unquotedResult,
+						core.ErrorSeverity.Information,
```
SPGoding: Why `Information`?

packages/core/src/parser/float.ts
SPGoding: How is Misode's point about JSON and mcdoc handled? For all other shared core parsers with configurable behavior we have those as a config option on the parser, and the checking is owned by the parser itself, not piggybacked within another package

packages/java-edition/src/mcfunction/checker/index.ts
MulverineX: I don't think we do? But does it matter in this case? mcdoc type literals are the same thing as snbt literals usually.

packages/java-edition/src/mcfunction/checker/index.ts
MulverineX: Not any particular reason, I'll move it.

packages/nbt/src/colorizer/index.ts
MulverineX: > Is the only purpose of this to create operator tokens for the colon and the comma?

Yes

> they are not operators so the token is wrong

What would be the correct token type? I'm not aware of any way to get the our LS to "clear" the color/token that was set by the Textmate.

> our semantic coloring is already coloring too much stuff; syntactical level stuff should be left to the much faster TextMate grammar on VS Code

syntax-mcfunction produces the incorrect color between an SNBT key and value if there is no space between the colon and value and the value starts with `[a-z9-0-_]`. Given the severe restrictions to Textmate I'm not sure its possible to fix, not an issue with my tree-sitter-mcfunction but VSCode is our primary platform. 

Additionally, I really don't think its that big of an issue to utilize semantic coloring like this.

packages/nbt/src/parser/primitive.ts
MulverineX: Because `1_000_000` is a perfectly valid unquoted string value pre-1.21.5. It therefore shouldn't be an error nor a warning IMO.

packages/core/src/parser/float.ts
MulverineX: Is there a reason not to support them in mcdoc? mcdoc type literals are supposed to basically be snbt right?

As for JSON, I'm not sure how its an issue there?

image

packages/nbt/src/parser/primitive.ts
MulverineX: A single value can match multiple NumeralPatterns in one loop pass. For example, the hex literal `0x1_2` matches both `IntegerAlike` and `LongAlike`. Without the flag the loop will emit `underscore-not-supported` when the only error you actually want is `radix-not-supported`. I can make it report both errors but that'd just be noise imo.

packages/java-edition/src/mcfunction/checker/index.ts
SPGoding: I didn't think we had int array literals in mcdoc. This is fine then if we don't have `snbt` highlights.

packages/nbt/src/parser/primitive.ts
SPGoding: That makes sense. Could you maybe add a comment to explain this? It wasn't obvious to me from glancing at the code

packages/nbt/src/colorizer/index.ts
SPGoding: That makes sense, thanks for the explanation. Maybe add a new `punctuation` token that maps to the TextMate scopes of punctuations (which I believe is separate from operotars)?

packages/java-edition/src/mcfunction/checker/index.ts
MulverineX: Apparently Misode's vscode-mcdoc just happens to support it by chance LOL. But hey it works, and yeah I just confirmed we don't have an SNBT textmate

packages/core/src/parser/float.ts
MulverineX: It turns out that this PR actually doesn't introduce support for underscore separators in `.mcdoc` anyway because of its own parser so this really doesn't matter IMO

SPGoding: > As for JSON, I'm not sure how its an issue there?

That error comes from the JSON language service. Spyglass suppresses its parser errors for JSON files, but I still don't think it should accept them under the hood.

MulverineX: Taken care of :+1: thanks

SPGoding (approved):
Thanks Mulv!

packages/java-edition/src/checker/nbtSyntax.ts
```diff
@@ -0,0 +1,152 @@
...
+ * checks that used to live inside the parser. The parsers now produce the
+ * new-syntax AST unconditionally; this step gates the result against the
+ * project's `loadedVersion`.
+ *
+ * Errors reported:
+ *
+ * - `snbt-functions-not-supported` - `bool(...)` / `uuid(...)` before 1.21.5.
+ * - `radix-not-supported` - hex/binary literals before 1.21.5.
+ * - `explicit-int-suffix-not-supported` - `42i` / `42I` before 1.21.5.
+ * - `underscore-not-supported` - digit separators before 1.21.5 (info).
```
NeunEinser: `(info)`? This should also be an error.

packages/java-edition/src/checker/nbtSyntax.ts
```diff
@@ -0,0 +1,152 @@
...
+		}
+	} else if (
+		NbtNumberNode.is(node) && node.hasUnderscoreSeparator && state.isOldSyntax
+		&& !state.underscoreNotified
+	) {
+		// `1_000_000` is a perfectly valid unquoted string value pre-1.21.5. Therefore, this shouldn't be an error nor a warning.
+		ctx.err.report(
+			localize('nbt.parser.number.underscore-not-supported'),
+			node,
+			core.ErrorSeverity.Information,
```
NeunEinser, line 56: Error

packages/java-edition/src/checker/nbtSyntax.ts
```diff
@@ -0,0 +1,152 @@
...
+
+interface WalkState {
+	isOldSyntax: boolean
+	/** Tracks whether the underscore-separator info has already been emitted. */
+	underscoreNotified: boolean
+}
+
+function walk(node: NbtNode, ctx: core.CheckerContext, state: WalkState): void {
+	if (NbtBoolFunctionNode.is(node) || NbtUuidFunctionNode.is(node)) {
+		if (state.isOldSyntax) {
```
NeunEinser: I really don't like this giant if else chain with repeated `if state.isOldSyntax`.

Could we maybe do:
1. Have `isOldSyntax` at the top level instead of repeated?
```ts
if (state.isOldSyntax) {
  // check nodes that have old syntax rules
} else {
  // check nodes that have new syntax rules
}
```
2. For the basic cases maybe have a constant map like `unsupportedSyntaxNodesToErrorMessage = {'nbt:hex': 'nbt.parser.number.radix-not-supported', 'nbt:bin': '...','nbt:bool_function': '...', 'nbt:uuid_function': '...' }`, so at least the simple cases of a node being completely unsupported can just be one line and easier to read?

packages/core/src/service/MetaRegistry.ts
```diff
@@ -67,6 +74,19 @@ interface SymbolRegistrarRegistration {
...
+	 * descends. Consumers (e.g. java-edition) can set this to gate
+	 * version-aware SNBT syntax features (hex/binary literals,
+	 * `bool(...)`/`uuid(...)` calls, underscore digit separators, ...) without
+	 * each call site having to thread the check through per-call options.
+	 *
+	 * The hook is expected to walk its argument recursively if it cares about
+	 * nested nodes. The signature is intentionally loose to avoid dragging
+	 * {@link @spyglassmc/nbt}'s types into {@link @spyglassmc/core}.
+	 */
+	snbtSyntaxCheck?: SnbtSyntaxCheck
```
NeunEinser: Isn't this just a normal checker? Why do we need this as a field somewhere instead of registering it as a normal checker?

mcfunction for instance also has some java-edition specific stuff which is just registered in the je package.

You can also do specific checkers for specific nodes, e.g. the je mcfunction checker has
```ts
	meta.registerChecker<mcf.McfunctionNode>('mcfunction:entry', entry)
	meta.registerChecker<mcf.CommandNode>('mcfunction:command', command)
	meta.registerChecker<BlockNode>('mcfunction:block', block)
  // ...
```

This could also resolve my comment/suggestion about the giant if/else chain.

Is there a reason this doesn't work?

In general I don't like polluting the core registry with a hardcoded dependency from another package.

packages/nbt/src/node/index.ts
```diff
@@ -93,6 +132,102 @@ export namespace NbtLongNode {
...
+}
+// #endregion
+
+// #region NbtSnbtFunctionNode
+// Base type for SNBT function calls (e.g. `bool(value)`). Concrete function node types
+// (e.g. `nbt:bool_function`) extend this. `prefixRange` covers the function name and opening
+// parenthesis (e.g. `bool(`); `suffixRange` covers the closing parenthesis (`)`). The
+// argument(s) are stored as `children`. Colorizers color `prefixRange` and `suffixRange`
+// as `escape`.
+export type NbtSnbtFunctionNode = NbtBoolFunctionNode | NbtUuidFunctionNode
```
NeunEinser: In theory all of there really are SnbtNodes. Imo you could rename everything into Snbt...Node (except for the NbtPathNodes), but I don't really know why this single type is NbtSnbtFunctionNode, especially when both other nodes are just Nbt(Bool|Uuid).

I think just calling this NbtFunctionNode would make more sense.

packages/nbt/src/parser/primitive.ts
```diff
@@ -130,6 +181,198 @@ export const primitive: core.InfallibleParser<NbtPrimitiveNode> = (
...
+			// Hex/binary literals: skip the prefix (`0x` or `0b`) in source, capture the
+			// remaining digits, then convert to bigint via BigInt. The radix is
+			// determined by the actual prefix character. If a type suffix
+			// (`b`/`s`/`i`/`I`/`l`/`L`/`f`/`d`) follows the digits we branch into
+			// the matching typed node; otherwise we keep the generic
+			// `nbt:hex`/`nbt:bin` BigInt for mcdoc/runtime to widen.
+			// Note: Has support for negative hex/binary input for future proofing,
+			// support for them is prevented by the regex and a specific error is
+			// provided.
+			if (e.group === Group.LongAlike && e.radix) {
```
NeunEinser, line 192: I don't fully understand why these new nodes are needed. Why can't we just use byte/short/int/long nodes (potentially with a boolean indicating whether it has a 0x/0b prefix?

Also the runtime checker should be passed the exact type, and not just a generic “this is some hex/bin literal”. It needs to know the exact NBT type.

Colorizer also seems to color the 0x/0b prefix in red. Not sure if intentional, but other languages typically just use normal number coloring for the entire literal, which imo makes more sense.

packages/nbt/src/node/index.ts
MulverineX: This was an intentional decision but now that I reconsider, yeah its a little silly

packages/java-edition/src/checker/nbtSyntax.ts
```diff
@@ -0,0 +1,176 @@
...
+			core.ErrorSeverity.Error,
+		)
+	}
+}
+
+export function register(meta: core.MetaRegistry): void {
+	meta.registerChecker<NbtBoolFunctionNode>('nbt:bool_function', checkBoolFunction)
+	meta.registerChecker<NbtUuidFunctionNode>('nbt:uuid_function', checkUuidFunction)
+	meta.registerChecker<NbtLongNode>('nbt:long', checkLong)
+	meta.registerChecker<NbtIntNode>('nbt:int', checkInt)
```
misode, line 140: Since there can (currently) only be one checker registered per AST type, these checkers should really be moved to the `nbt` package.

I know it runs into the problem of the `nbt` package needing to check the java-edition loadedVersion, but that's a problem we already had before so we can fix that in the future.

packages/java-edition/src/checker/nbtSyntax.ts
```diff
@@ -0,0 +1,176 @@
...
+
+/** Minimum game version that supports the SNBT additions gated by this step. */
+const MIN_NEW_SYNTAX: ReleaseVersion = '1.21.5'
+
+/**
+ * Per-checker-call dedup for the underscore-separator info-level
+ * diagnostic. The walker used to thread a `state` object through the
+ * recursion; without that, a module-level `WeakSet` keyed on the
+ * checker context does the same job - and naturally GC's when the
+ * context is dropped.
```
misode: These comments (and others) mention "walker" and "state", pretty sure that code was refactored at some point. It also reads really unnatural. I suggest either removing a lot of these comments or rewriting them to use simpler language and removing unnecessary information.

packages/java-edition/src/checker/nbtSyntax.ts
MulverineX: discussed on discord, SPGoding and Neun are okay with doing the hack for now

packages/java-edition/src/checker/nbtSyntax.ts
```diff
@@ -15,16 +15,9 @@ import type {
...
 
-/**
- * Per-checker-call dedup for the underscore-separator info-level
- * diagnostic. The walker used to thread a `state` object through the
- * recursion; without that, a module-level `WeakSet` keyed on the
- * checker context does the same job - and naturally GC's when the
- * context is dropped.
- */
+// Report `underscore-not-supported` at most once per checker call.
 const underscoreNotifiedContexts = new WeakSet<core.CheckerContext>()
```
misode, line 21: Why the need for a global variable like this? This looks very janky imo.

packages/java-edition/src/index.ts
```diff
@@ -153,5 +154,11 @@ export const initialize: core.ProjectInitializer = async (ctx) => {
 	jeMcf.initialize(ctx, summary.commands, release)
 	nbt.initialize(ctx)
 
+	// Register per-NbtNode-type version-aware SNBT-syntax checkers. The
+	// core checker framework dispatches them by node type during descent,
+	// so both mcfunction and JSON-via-mcdoc-string paths pick them up
+	// without each call site having to thread an extra check option.
```
misode, line 160: This is just the LLM repeating your refactor prompt, totally unnecessary comment because this is how all checkers work.

PR #2073, opened 2026-08-15
🐛 Fix json number conversion allowing to convert floats to ints

NeunEinser:


packages/mcdoc/src/runtime/checker/error.ts
```diff
@@ -547,7 +547,7 @@ export function getDefaultErrorReporter<T>(
 								)
 								: e.kind === 'literal'
 								? localeQuote(e.value.value.toString())
-								: localize(`mcdoc.type.${e.kind}`)
+								: localize(`mcdoc.type.${e.kind === 'tuple' ? 'list' : e.kind}`)
```
misode, line 550: Why don't we just add a translation key for tuple?

NeunEinser: I didn't want to do that, because “tuple” is more implementation detail in this context. But if we were to show the type in an mcdoc file somewhere, it should say “tuple”, not list.

So basically want to keep the option open to translate it as “tuple”.

MulverineX (approved):

SPGoding (approved):

PR #2093, opened 2026-09-12
🐛 Fix mcdoc resolving to simplified form too early

### LINKED ISSUE #1537: Nested mcdoc spread operator failing to appear at runtime
* Source code:

The complete code can be found at [ui.mcdoc](https://github.com/Alumopper/Floating-UI/blob/master/mcdoc/ui.mcdoc)

```rust
struct UI {
    type: #[dispatcher_key="floating_ui:ui"] string,
    name?: string,
    tag?: string,
    ...floating_ui:ui[[type]]
}

struct Control {
    x?: double,
    y?: double,
    size?: [float]@2,
    rotation?: [float]@4,
    display?: enum(string){
        none = "none",
        thirdperson_lefthand = "thirdperson_lefthand",
        thirdperson_righthand = "thirdperson_righthand",
        firstperson_lefthand = "firstperson_lefthand",
        firstperson_righthand = "firstperson_righthand",
        head = "head",
        gui = "gui",
        ground = "ground",
        fixed = "fixed"
    },
    item?: Item,
    move_in?: string,
    move_out?: string,
    anim?: ControlAnimate,
    children?: [int[]@4]
}

dispatch floating_ui:ui[panel] to struct Panel {
    ...Control,
    child: [Control]
}
```

* What I expected:

When I type `data modify storage floating_ui:input data set value {“type”:"panel",}`, there should be suggestion for members defined in struct `Panel`

{A9D6F85C-C2FE-4878-8D0A-49B91D1F167A}

* What happens actually:

There are only members defined in struct `UI` can be auto-complete: 

{04F04278-6BB6-4966-974C-EEDB1CBDB2B2}

### LINKED ISSUE #1990: The left_rotation and right_rotation of the display entity does not accept axis angle format
## Source code:

```mcfunction
data modify entity @n[tag=hook,tag=!chu_shi_hua_le] transformation.left_rotation set value {axis: [0, 1, 0], angle: 1.5707963267948966192313216916398}
```

Image

Image

## What I expected:

The `left_rotation` and `right_rotation` of the display entity accept axis angle format

## What happens actually:

The `left_rotation` and `right_rotation` of the display entity does not accept axis angle format

### LINKED ISSUE #1996: Completely missing auto-completion directly after compound filter in nbt paths
If you write the following command and place the cursor at the end, there is no auto-completion for any of the fields:
`data get entity @s {}.`
It works fine if there is already another field between the compound and the cursor like in
`data get entity @s {}.Item.`
This only applies to compound filters, not list accesses.

Dahesor:
Fixes #1996
Fixes #1990 
Realized that this also fixes #1537 

Been watching this issue for a long time, decided to give it a try.

After some digging, Both #1990 and #1537 has the same cause, that before getting to the leaf node, the `McdocType` is already simplified down to the `SimplifiedMcdocType` in the context of the NBT path.

The idea here is to also save the original McdocType as `endOriginalTypeDef` so it can be used later. Not sure if this is the best way but it worked locally.

#1996 is arguably not related to the other 2 issues. The cause is Spyglass ignored the filter, and resulted in its `path.endTypeDef` to become void. The solution is simply have the filter's type info attached as its `link.next`.


packages/mcdoc/src/runtime/checker/index.ts
```diff
@@ -377,10 +382,20 @@ function attachTypeInfo<T>(node: CheckerTreeRuntimeNode<T>, ctx: McdocCheckerCon
...
+			{
+				kind: 'union',
+				members: definitions.map(d => d.typeDef),
+			},
+			undefined,
+			{
+				kind: 'union',
+				members: node.definitionsByParent
+					.filter(d => d.validDefinitions.length > 0)
+					.map(d => d.originalTypeDef),
```
NeunEinser, line 396: Oh neat, I didn't think accessing the original typeDef was so easy, but yeah apparently we already keep it in the tree anyways :D

NeunEinser (approved):
Looks good, makes sense, and is suprisingly simple! 🎉

MulverineX (approved):

MulverineX (approved):

MulverineX:
*Thanks vscode lol*

PR #2094, opened 2026-09-13
🏗️ Allow registering multiple checkers of the same AST type

misode:
Allows calling `registerChecker` with the same type multiple times. Instead of overwriting the checker, now chains the two checkers after each other. This is maybe also applicable to binders, but apart from that not really to the other processors.

Important before merging: We need to absolutely make sure that the project initializers are only ever called once for a given `MetaRegistry`, otherwise we'll get double of all checkers.


NeunEinser (approved):

MulverineX:
Waiting on rigorous testing & review from SPG (misode's intent from the discord) since this is a bit tough to get right.

Also since this is probably going to be merged after my PRs maybe the refactor should happen here.

packages/core/src/service/MetaRegistry.ts
```diff
@@ -155,7 +155,15 @@ export class MetaRegistry {
 		return this.#checkers.get(type) ?? checker.fallback
 	}
 	public registerChecker<N extends AstNode>(type: N['type'], checker: Checker<N>): void {
-		this.#checkers.set(type, checker)
+		const oldChecker = this.#checkers.get(type)
+		if (oldChecker !== undefined) {
+			this.#checkers.set(type, (node, ctx) => {
```
SPGoding, line 160: This is quite smart, but what do you think about changing `this.#checkers` to be a Map of Sets instead? We'll need to change more calling sites but I think that prevents the issue of storing duplicate checkers.

misode: Given that the order of `Set` is guaranteed in JS, I think that's a good solution yes.

PR #2095, opened 2026-09-14
✨ Make mcdoc/runtime see numbers with underscore as string in old MC versions

NeunEinser:


MulverineX:
We ended up deciding to not do this, opting instead for making them warnings as a compromise

Issue #1795, opened 2025-04-16
NBT Path Completion Does Not Contain All Possible Values When There Are Multiple Cases


Dahesor:
For example:

![Image](https://github.com/user-attachments/assets/00e15448-104e-4f65-91a6-95d10c487ea9)
When writing SNBT, which keys are available is determined by the value of `action`. However when writing a NBT path Spyglass should suggest all the possible keys (like `command`, `url`, etc.), not just the one that determine the case.

Thanks

NeunEinser:
Mmh, yeah that's unfortunately not very easy with how things are implemented.

Basically, `action` is usually used to determine the type of click event, so if you add a click event, you would first specify action and then get auto-complete for the other fields.

Obviously, that doesn't work for nbt paths.

You'll get the same issue if you do sth like `Passengers[].`, you will only get the `id` completion here.

PR #2097, opened 2026-09-15
🔨 Add mcdoc checking to NBT path filters

### LINKED ISSUE #1365: Provide mcdoc/runtime in index and filter segments of nbt path
Provide contextual error checking and auto-complete in indexed/filtered segments of nbt paths

Dahesor:
Fixes #1365 
Is a workaround for #1795 

This adds contextual error checking and auto-complete for NBT path filters `path.object{...}` and `path.list[{...}]`.

When a filter is present, it will use the keys in the filter as the dispatcher key for the completion of the following NBT path segments, if available.
<details>
<summary>See behaviour</summary>

- `text.click_event`  -> suggests only `action`
- `text.click_event{action:"run_command"}` -> suggests both `action` and `command`
- `text.click_event{action:"open_url"}` -> suggests both `action` and `url`
- `Passengers[{id:"zombie"}]` -> suggests zombie nbt and `id`
- `text{click_event:{action:"run_command"}}.click_event`  -> suggests both `action` and `command`
- `text[{click_event:{action:"run_command"}}].click_event`  -> suggests both `action` and `command`
**However**:
- `text.click_event{command:"say 1"}` -> suggests only `action`
- `text{click_event:{action:"run_command"}}.click_event{command:"say 1"}`  -> suggests only `action`, since the data is overritten by the later filter.
- `Passengers[0]` -> suggests only id. And the work around doesn't work since we can't have both an index and a filter.

</details>

The idea is, an optional `filter_data` is attached to NbtPathLink, which stays undefined until a filter node. At each node that is not a filter, it will try to find itself from the keys in the filter passed by the parent, if it can then it will pass that protion to its children.
A `getChildrenFromFilterData` is added besides the original `getChildren`, and we will use the first one if available.


MulverineX:
Going to test this in editor before approving, looks good though!

MulverineX (approved):

PR #2101, opened 2026-09-16
🐛 Fix #2096

### LINKED ISSUE #2096: When there is a union of structs, fields still retain type information from structs that have missing key errors
Minimum Example:
```rust
type Baz = (
  struct { foo: int } |
  struct { bar: string, [any]: any } |
)

NeunEinser:
Fix #2096


NeunEinser:
For reviewing the actual fix, just look at the last commit separetely, as it also contains relevant snapshot changes.

The other commits are refactoring the runtime test setup a lot.

